PR #1, opened 2024-08-27
CMS-356: Add directory structure, dev containers

duncan-oxd:
Using patterns (but not code) from other recent OXD projects, I set up a basic project structure and (mostly) configured dev containers.

There are some things left to do, but I want to get this reviewed to discuss and merge it so that we can start writing code for any part of the project.

The dir structure is simply `frontend/`, `backend/`, and some configuration files in the root. `db/` just has an init script to populate the fake DB with some example rows.

`docker-compose.prod.yml` uses the `Dockerfile.prod` files in each subfolder, plus a placeholder Postgres DB to spin up a demo app for testing. We can change anything there but I thought it might help for testing/developing the Openshift setup.

`docker-compose.dev.yml` is used for the VSCode dev containers. (Apparently dev containers work in some other IDEs but I've never tried). If you open the root folder in VSCode, it should recommend installing the Dev Containers extension if you don't already have it. Then you can "reopen in a container" and select Frontend or Backend. You can open a new window to run both at the same time, that was a challenge to configure. I'll document all of this in the Readme once it's finalized. Right now the only sort of bug I've noticed is that git doesn't work inside the containers. I found a GitHub bug issue with a workaround, but I haven't tried it yet.

Check it out and let me know what needs to change!


docker-compose.prod.yml
```diff
@@ -0,0 +1,35 @@
+services:
```
molund, line 1: we might not end up using this file, let's keep it just in case.

backend/index.js
```diff
@@ -0,0 +1,27 @@
+import http from "http";
```
molund, line 1: This is a good start, but we'll soon need an ORM and something to handle database migrations

molund (approved):
This is a good start, especially the dev containers work.  I'll want to flesh out the express app quite a bit, but it will be good to get this merged so we can both start working on it.

ayumi-oxd:
It'd be nice if there's a PR template something like this: https://github.com/bcgov/bcparks.ca/blob/main/.github/pull_request_template.md, but not an urgent thing 🙂

backend/index.js
duncan-oxd: I'd like to split that into a separate ticket - and we'll use something like Express (or Fastify?) instead of a basic node server 👍 

Mostly wanted to get the base boilerplate stuff set up so we'll all have a common starting point for adding to it. I'm about to push some Eslint tweaks, then I'll merge it and we can create some specific tickets to tackle

duncan-oxd:
I know there are a ton of commits, I'll be sure to squash them 😆 
There was a lot of trial and error with the dev containers, oof.

docker-compose.prod.yml
duncan-oxd: I figured it was just a good reference for the services and the ports they use (and if you want to test it locally). Let's remember to delete it if it becomes out of date :P

PR #2, opened 2024-08-29
CMS-356: Install Fastify and Sequelize

duncan-oxd:
### Jira Ticket

CMS-356

### Description

Installed the **Fastify** web framework and set up some example routes to test it out, plus the **Sequelize** ORM.

ExpressJS is still the most popular, but it has more or less been in maintenance mode for years, and a lot of the core team is now working on Fastify. In recent years, Fastify is one of the big ones that people recommend. Nicer syntax, and better support for types and testing frameworks, if we ever want to add those. Also better performance, but I doubt that would be noticeable :P It has a slick auto-loader to load routes and plugins without needing to manually import everything.
I'm not concerned about either one becoming unsupported any time soon, but we can easily switch over to Express if we want.

Sequelize is the most popular ORM for NodeJS. I think it should be fine for anything we need to do on this project. I defined a model and made a demo endpoint to retrieve rows from the example table:

http://localhost:8000/nested-path-example/orm-1
http://localhost:8000/nested-path-example/orm-2

### DIFF
--- a/backend/routes/nested-path-example/db-row.js
+++ b/backend/routes/nested-path-example/db-row.js
@@ -2,10 +2,11 @@ import pgp from "pg-promise";
 
 const db = pgp()(process.env.PG_CONNECTION_STRING);
 
+// example with a manual query
+// http://0.0.0.0:8000/nested-path-example/row-1
 // http://0.0.0.0:8000/nested-path-example/row-2
-// http://0.0.0.0:8000/nested-path-example/row-2
-export default async function (fastify, opts) {
-  fastify.get("/row-:rowId", async function (request, reply) {
+export default async function (fastify) {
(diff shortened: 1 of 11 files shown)


backend/routes/nested-path-example/db-row.js
```diff
@@ -0,0 +1,19 @@
+import pgp from "pg-promise";
+
+const db = pgp()(process.env.PG_CONNECTION_STRING);
```
duncan-oxd, line 3: I left this in as an example, if we need to do any manual queries without the ORM. We probably won't use this, so we can take it out when we start adding the real routes

duncan-oxd:
Closing this one in favor of another branch with Express instead of Fastify, coming up shortly... 👍

PR #7, opened 2024-09-05
CMS-382: Build header and footer in MainLayout

duncan-oxd:
### Jira Ticket

CMS-382

### Description

Installed the BC Parks bootstrap theme from digitalspace

Built the basic header and footer in the main layout file, based on the Attendance and Revenue reference that Amanda provided. It's pretty close to the A&R layout, but not pixel perfect. The A&R layout uses some overly complicated CSS and I'd kinda prefer to have it a couple of pixels off rather than copy what they've done.

As of yet, I think we're waiting to hear back about which email address to use for the "Contact us" footer link. I'll update that but the rest is ready to look at.
- [ ] Check with Amanda about the Contact email


frontend/src/router/layouts/MainLayout.scss
```diff
@@ -0,0 +1,27 @@
...
+  & > .container {
+    flex: 1;
+  }
+
+  & > footer {
+    min-height: 4.1rem;
+
+    .quick-links {
+      gap: 0 1.5em;
+
```
ayumi-oxd, line 15: Just my personal preference but I'm not a big fan of mixing `rem` and `em`, but I would love to hear if there's a reason to use them together. e.g. `.quick-links` gap is based on its parental element.

frontend/src/router/layouts/MainLayout.scss
```diff
@@ -0,0 +1,27 @@
...
+      a {
+        color: var(--bs-white);
+        text-decoration: none;
```
ayumi-oxd, line 18: Just curious to know what made you to use `var()` instead of variables in SCSS.

ayumi-oxd (approved):
I left some nitpicky questions, but it looks good to me 👏

frontend/src/router/layouts/MainLayout.scss
duncan-oxd: Great question, and I don't have a strong reason in this particular case but I'll explain my idea:

If we change the base font size, then everything scales proportionally:

image

If we just change the font size of the footer element, then the blue box stays the same height, but the font size changes:

image

I figure the height of the blue box (in `rem`) is a layout element and we'd prefer if it stays roughly that same height, even if the font inside changes. But he spacing between the links (`em`) should scale with the font.

Here's an exaggerated version to demonstrate:

image

In this case it barely makes a difference, but sometimes it can have an impact. (If users zoom in the page like my mom does, or if they change their screen text scaling in their accessibility settings, etc). Mostly so we can control what shrinks and grows in an element if we change its font size.

Let me know if that makes sense! It doesn't do much in this case so I could change it if you'd prefer

frontend/src/router/layouts/MainLayout.scss
duncan-oxd: I'm trying to un-learn the habit of using SCSS vars for things like this when CSS variables are available. At the end of the day in our site it's going to be the same, but CSS variables let us change things at runtime: If the user switches to a dark theme, or if the client wants us to fade out the colors when we mouse over or something, we would just use JS to change a couple of variables, while we might need to override a lot more if the values were "hardcoded" from compiled SCSS.

duncan-oxd: ...but I haven't used Bootstrap in a long time and I'm pretty new to styling React components, so let me know if there's a better way to do things! CSS Variables didn't exist last time I used Bootstrap 😆

frontend/src/router/layouts/MainLayout.scss
ayumi-oxd: It does make sense! Thanks for your explanation 👏

duncan-oxd:
I just updated the email address from the comments in the ticket. This is failing the strapi/gatsby test but those don't apply here so I'll merge 👍

PR #9, opened 2024-09-11
CMS-321: Set up Keycloak

duncan-oxd:
### Jira Ticket

CMS-321 has many parts; this branch covers the "Setting up keycloak" part (as well as "Log into the portal with IDIR" & "Gains access to all areas of the Web Staff Portal through one login," more or less!)

### Description

Installed Keycloak packages on the frontend and backend:

#### Frontend

The "Foo" example route has some content protected by a login. (With log in/log out buttons)
It uses the same login realm as the Staff portal, so hopefully your session would persist across apps (once they're on the same domain). I haven't tried this yet, but if you're not logged in yet and you log in on this demo app and then navigate to the staff protal, you will be automatically logged in. We might need to check cookies to make it more seamless later, though.

If this all looks good, I'll remove the expicit login/logout buttons and protect the entire React app (or select routes) in another branch, because the whole thing will be behind a login in production.

#### Backend 

I put a couple of the API endpoints behind Keycloak as an example. You can click the button with the "🔒" emoji in the React app to test it.

#### Both

The Keycloak details are set with environment variables. It sounds like BC Parks might have a special way of handling environment variables so I can also create a branch to implement that if necessary 👍


backend/routes/nested-path-example/orm.js
```diff
@@ -1,28 +1,50 @@
...
+  res.json({ rowId, specificRow, numRows: allRows.length });
+});
+
+// example with Sequelize ORM 🔒 protected by Keycloak
+// http://0.0.0.0:8100/nested-path-example/orm-2
+router.get(
+  "/orm-2",
+
+  // protect this route with JWT authentication
+  checkJwt,
```
molund, line 27: I'm sure there's some magic going on here, but I could proably use a walk-through of how this actually works.

duncan-oxd: Definitely, I can go over it in more detail but the short version is that basically everything in Express is "middleware" and it all gets processed in order for each request. The "router" definitions are also middleware technically. So the `checkJwt` function basically runs before the actual URL handler function, and it verifies your token before anything else is allowed to happen.

So the order of middleware here is like:
- CORs, logging, other stuff defined in index.js
- Route handler stuff (see below)
- Finally there's an error handler, also in index.js, if the Route handler failed

The signature for `express.use` is like:
(`url match pattern, middleware function[, middleware function, ...]`)

For a route definition, the last function in the chain is the actual handler that returns to the browser, but anything you put before that gets run first. So the `checkJwt` function checks your access token and cancels it with a 401 error if the token is invalid.

duncan-oxd: ...and on that note, I think we'll use a similar middleware pattern to check the specific roles

1. token valid?
2. can edit data?
3. can edit THIS data?
4. if so, then run the handler function

molund (approved):

duncan-oxd:
Rebased, tested that it's all working on the /v2/ URL path ✅ Merging

PR #10, opened 2024-09-12
CMS-321: Protect entire routes with Keycloak

duncan-oxd:
### Jira Ticket

CMS-321

### Description

Expanding on #9 , this changes the page-specific and section-specific auth to protect the entire frontend app, and all the endpoints on the backend (except`/` and the `/time` testing endpoint in the root).

On the frontend, there's no explicit log in/out buttons. If you're not logged in, it will forward you to Keycloak instead of showing anything.

On the backend, it will continue rejecting requests with `401` if you don't send a valid token.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -39,9 +39,11 @@ const limiter = RateLimit({
 
 app.use(limiter);
 
-// add routes
+// Public routes
 app.use("/", homeRoutes); // example stuff for testing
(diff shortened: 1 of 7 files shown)


molund:
Note that we'll eventually have an anonymously available public API so more of the backend routes will need to be unprotected

molund (approved):

duncan-oxd:
Just rebased onto main; merging ✅

PR #11, opened 2024-09-16
CMS-407: Add entrypoint script, env.js

duncan-oxd:
### Jira Ticket

CMS-407

### Description

This adds an env.js file and an entrypoint script, similar to the one in the bcparks.ca "admin" directory, but it puts all of the variables in `window.envVars` instead of just `window` for peace of mind.

I moved the "/v2/" base setting into the vite config file, so it can be shared by the dev server as well as the build config. Some further config (or rewrite rules) might be required for it to work in production, depending on how we serve it.

If this looks good, I made another ticket, CMS-408, to use dynamic URLs instead of environment variables for the login/logout URLs.


frontend/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+echo "window.envVars = {};" > $APP_ROOT/env.js
+
+if [ -n "$VITE_OIDC_AUTHORITY" ]; then
+  echo "window.envVars.VITE_OIDC_AUTHORITY = \"$VITE_OIDC_AUTHORITY\";" >> $APP_ROOT/env.js
```
molund, line 6: these lines should actually write to `$APP_ROOT/v2/env.js` (the v2 app is in a subfolder called v2)

molund:
I think you'll need to add an empty env.js file in `/frontend/public` so users don't get a 404 when developing locally.

frontend/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
molund, line 1: We can't execute both entrypoint files when the pods start up.  
I would be okay with either of these options:
A) the frontend-legacy and frontend share the same entrypoint file
B) one script calls the other and then resumes its regular execution

frontend/.gitignore
```diff
@@ -1,3 +1,5 @@
+public/env.js
```
molund, line 1: We should server an empty file instead of ignoring it. See my general comment.

frontend/entrypoint.sh
duncan-oxd: I just pushed an update to combine the logic into the legacy frontend's existing entrypoint script. We'll just need a `V2_APP_ROOT` var to point to the frontend/public directory.

If option B makes more sense for how we're building and deploying, just let me know and I can update it 👍

molund (approved):
Looks good. I'm going to merge this so I can see if it works

PR #12, opened 2024-10-02
CMS-416: Clear session storage on log out

duncan-oxd:
### Jira Ticket

CMS-416

### Description

Now that the new app is running alongside the old Staff Portal sections on the same domain, we noticed that logging out of the old portal won't log you out of the new one. (The new one uses SessionStorage and the old one uses cookies and possibly LocalStorage). This branch is a quick fix to clear SessionStorage on the home page when you're logged out. 

A more comprehensive solution would be to make both sections use the same auth provider, and share the same storage location and tokens. TBD, but this might be worth doing to avoid possible hassles down the road.


ayumi-oxd (approved):
LGTM 👍

PR #14, opened 2024-10-04
CMS-422: Set up site structure

duncan-oxd:
### Jira Ticket

CMS-422

### Description

This branch creates the routes for all of the expected pages (as far as we know at the moment. A few more TBD may be added)

Also installs the Design Tokens package, even if we don't need it that much with the Bootstrap theme doing most of the same stuff.

I kept the "API testing route" and renamed it from "App" - I figure it's handy to keep around until we're settled on the ORM and the AdminJS stuff.


ayumi-oxd (approved):
LGTM 👍

PR #18, opened 2024-10-17
CMS-424: Build park detail layout

duncan-oxd:
### Jira Ticket

CMS-424

### Description

Built the "Park detail" page elements in bootstrap. After we set up the API endpoints, then we can hook these pages up to the data and turn this into React components.

An example URL for testing locally is:
http://localhost:8101/v2/park/abc99
Everything is static except the "park ID" which is printed out at the top for now.


frontend/src/router/pages/PageDetails.scss
```diff
@@ -0,0 +1,103 @@
...
+    display: flex;
+    align-items: center;
+    padding: var(--layout-padding-medium);
+
+    // seamless borders on larger screens
+    @media (min-width: 576px) {
+      border-left: none;
+      border-radius: 0 var(--layout-border-radius-medium)
+        var(--layout-border-radius-medium) 0;
+    }
```
duncan-oxd, line 94: I couldn't find an elegant way to use Bootstrap's breakpoint mixins with the BC Parks bootstrap theme. Let me know if you have any ideas. Otherwise, I'm fine with hardcoding it a few times for now.

ayumi-oxd (approved):
LGTM 👍 If you're tired of grabbing data manually, these could be temp options.

Protected areas: [https://cms.bcparks.ca/api/protected-areas?populate=*](https://cms.bcparks.ca/api/protected-areas?populate=*)
Elk Falls Park: [https://cms.bcparks.ca/api/protected-areas/28?populate=* ](https://cms.bcparks.ca/api/protected-areas/28?populate=* ) 
(^ Add park orcs number as an endpoint)

PR #17, opened 2024-10-16
CMS-423 Build landing page layout

duncan-oxd:
### Jira Ticket

CMS-423

### Description

Updated the layout on the "Dates management" landing page. The filter controls and the table are all still non-functional at this point, and haven't been abstracted into React components.

I installed font awesome and did some minor tweaks to the main layout.

PR #19, opened 2024-10-18
CMS-426: Build layout for Submit dates page

duncan-oxd:
### Jira Ticket

CMS-426

### Description

Built and styled the elements for the "Submit dates" (aka Edit dates) page. Similar to #18 and #17 it's just the html and CSS. We'll hook it up to the API data and make working components in a separate ticket.

- I wrapped the other page(s) style sheets in a selector to limit their scope to the one page ("Hide whitespace" in the PR diff to make it readable).  If there's a more elegant way to scope a stylesheet to one component/page, please let me know! (css modules?)
- An example URL for local testing is http://localhost:8101/v2/park/foo/edit/bar-baz (`/park/<park ID>/edit/<season ID>`)- I'm now thinking we might want to put the `/edit` at the end, but it may not make a difference here. 
- I'm trying to get it to match the wireframes as closely as possible while using built-in Bootstrap stuff, without writing too much custom code at this point (as per Mike's project goals) - It looks like they may want a full-width layout and we can make a ticket to implement that if it's definite.

### DIFF
--- a/frontend/src/router/pages/SubmitDates.jsx
+++ b/frontend/src/router/pages/SubmitDates.jsx
@@ -94,7 +94,7 @@ function SubmitDates() {
             </div>
 
             <p>
-              <button className="btn btn-text-primary">
+              <button className="btn btn-text-primary text-link">
                 + Add more operating dates
               </button>
             </p>
@@ -146,7 +146,7 @@ function SubmitDates() {
             </div>
 
             <p>
-              <button className="btn btn-text-primary">
+              <button className="btn btn-text-primary text-link">
                 + Add more reservation dates
               </button>
             </p>
--- a/frontend/src/router/pages/SubmitDates.jsx
+++ b/frontend/src/router/pages/SubmitDates.jsx
@@ -1,0 +1,230 @@
+import { useParams } from "react-router-dom";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import groupCamping from "../../assets/icons/group-camping.svg";
+
+import "./SubmitDates.scss";
+
+function SubmitDates() {
+  // @TODO: fetch park name + details
+  const { parkId, seasonId } = useParams();
+
+  function SubArea() {
+    return (
+      <section className="sub-area">
+        <h3>Alouette groupsite</h3>
+
+        <div className="row">
+          <div className="col-md-6">
+            <p>
+              Operating dates{" "}
+              <FontAwesomeIcon
+                className="append-content"
+                icon="fa-solid fa-circle-info"
+              />
+            </p>
+
+            <p>Previous: Mon, Jun 7, 2025 - Wed, Sep 15, 2025</p>
+
+            <div className="row dates-row operating-dates">
+              <div className="col-lg-5">
+                <div className="form-group">
+                  <label htmlFor="startDate0" className="form-label d-lg-none">
+                    Start date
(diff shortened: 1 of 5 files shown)


frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -0,0 +1,230 @@
...
+        <h3>Alouette groupsite</h3>
+
+        <div className="row">
+          <div className="col-md-6">
+            <p>
+              Operating dates{" "}
+              <FontAwesomeIcon
+                className="append-content"
+                icon="fa-solid fa-circle-info"
+              />
```
ayumi-oxd, line 23: This icon in the mock seems to be a regular one (white background). Also, we need to add some text to these tooltips at some point.

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -0,0 +1,230 @@
...
+      </header>
+
+      <h2>
+        <img src={groupCamping} className="sub-area-icon" /> Group Camping
+      </h2>
+
+      <p className="mb-5">
+        <a href="#TODO:-holidays-link">View a list of all statutory holidays</a>
+      </p>
+
```
ayumi-oxd, line 174: Just FYI, we use this API for the BC Parks: [https://canada-holidays.ca/api/v1/provinces/BC](https://canada-holidays.ca/api/v1/provinces/BC)

ayumi-oxd (approved):
LGTM! I kinda prefer to have `/edit` at the end but either way should be fine 👍

frontend/src/router/pages/SubmitDates.jsx
ayumi-oxd: Oh I guess it can be just a link.

duncan-oxd: I'll have to see what they have in mind for that link. (If it pops up a modal window with our own list, or if it just links to some other page...) I'll leave it as a "TODO" for now but I don't want to forget it!

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -0,0 +1,230 @@
...
+                  />
+                </div>
+              </div>
+            </div>
+
+            <p>
+              <button className="btn btn-text-primary">
+                + Add more operating dates
+              </button>
+            </p>
```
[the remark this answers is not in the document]
duncan-oxd: Good catch, some buttons that look like links, but with no underlines 👀

duncan-oxd:
I'll make myself a small ticket in the backlog to update the URL paths. Maybe:
/park/:parkId/:seasonId/edit
/park/:parkId/:seasonId/review
❓ 🤔 

It depends on if there will be more than just the seasons under the "park" part of the path. Right now it looks like it's just parks->seasons so the structure above could work.

frontend/src/router/pages/SubmitDates.jsx
duncan-oxd: Yeah 😮‍💨 a mix of two different icon sets.. I'm dreading setting up a "Pro icon kit" for FontAwesome here but I think that's what we'll have to do. When we build out the actual React components, we can use the react-bootstrap tooltip component ✅ 

I'll leave the icon with a placeholder until we sort out the tooltip content 👍

PR #20, opened 2024-10-22
CMS-425: Add Publish and Export pages

duncan-oxd:
### Jira Ticket

CMS-425

### Description

This branch builds out the elements for the "Publish" and "Export" pages. I moved them in the router config so they're children of the "/" landing page route (and they'll share a tabbed navigation with the "Dates management" table page).

I also removed the table row hover styles from all tables. It's not in the wireframes and I don't want to suggest the row is clickable when it's not.

The layout still uses the built-in Bootstrap container class, even though the wireframes are full width for large screens. I'll discuss with design and see what the final decision is for the responsive layout in another ticket.

### DIFF
--- a/frontend/src/router/layouts/LandingPageTabs.jsx
+++ b/frontend/src/router/layouts/LandingPageTabs.jsx
@@ -14,14 +14,13 @@ export default function LandingPageTabs() {
             </NavLink>
           </li>
           <li className="nav-item">
-            <NavLink className="nav-link" to="/export">
-              Export
-            </NavLink>
-          </li>
-          <li className="nav-item"></li>
-          <li className="nav-item">
             <NavLink className="nav-link" to="/publish">
               Publish
+            </NavLink>
+          </li>
+          <li className="nav-item">
+            <NavLink className="nav-link" to="/export">
+              Export
             </NavLink>
           </li>
         </ul>
--- a/frontend/src/router/layouts/LandingPageTabs.jsx
+++ b/frontend/src/router/layouts/LandingPageTabs.jsx
@@ -1,0 +1,34 @@
+import { Outlet, NavLink } from "react-router-dom";
+import "./LandingPageTabs.scss";
+
+export default function LandingPageTabs() {
+  return (
+    <div className="layout landing-page-tabs">
+      <header className="section-tabs d-flex flex-column">
+        <h1 className="m-3 mb-4">Dates management</h1>
+
+        <ul className="nav nav-tabs px-2">
+          <li className="nav-item">
+            <NavLink className="nav-link" to="/">
+              Edit and Review
+            </NavLink>
+          </li>
+          <li className="nav-item">
+            <NavLink className="nav-link" to="/export">
+              Export
+            </NavLink>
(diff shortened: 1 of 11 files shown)


frontend/src/router/pages/ExportPage.jsx
```diff
@@ -0,0 +1,161 @@
...
+                className="form-check-input"
+                type="radio"
+                name="type"
+                id="type-all-dates"
+                value="all-dates"
+              />
+              <label className="form-check-label" htmlFor="type-all-dates">
+                All dates
+              </label>
+            </div>
```
ayumi-oxd, line 24: All labels looked smaller than the mock (12px). Should it be 16px?

frontend/src/router/layouts/LandingPageTabs.jsx
```diff
@@ -0,0 +1,34 @@
...
+            <NavLink className="nav-link" to="/export">
+              Export
+            </NavLink>
+          </li>
+          <li className="nav-item"></li>
+          <li className="nav-item">
+            <NavLink className="nav-link" to="/publish">
+              Publish
+            </NavLink>
+          </li>
```
ayumi-oxd, line 26: "Publish" comes second and "Export" comes third in the mock.

frontend/src/router/pages/ExportPage.jsx
```diff
@@ -0,0 +1,161 @@
...
+          <fieldset className="mb-3">
+            <legend className="append-required">Year</legend>
+
+            <div className="input-with-append">
+              <select id="year" className="form-select"></select>
+              <FontAwesomeIcon
+                className="append-content"
+                icon="fa-solid fa-magnifying-glass"
+              />
+            </div>
```
ayumi-oxd, line 49: Should it be a calendar icon?

ayumi-oxd (approved):
LGTM! I left some nitpicky questions but overall looks great 👍

frontend/src/router/pages/ExportPage.jsx
duncan-oxd: Oops, right! I'll fix that.

frontend/src/router/pages/ExportPage.jsx
duncan-oxd: It does look small, but that's how looks is in the bootstrap theme 🤷 
https://digitalspace.github.io/bcparks-bootstrap-theme/#forms

My understanding is that the Figma link is just wireframes and not an exact design mockup so I've just been using the default Bootstrap styles as much as possible.. But I don't think anything should be less than 16px in the app. I'll add an override to make the radio/checkbox label font bigger and check with design to see if there's a specific idea.

duncan-oxd: Oh, I see the "label" size is supposed to be 0.75rem in the design tokens: https://www2.gov.bc.ca/gov/content/digital/design-system/foundations/design-tokens/glossary
But that is probably meant for text box labels and not checkbox and radio buttons. I'll make an override to leave those ones at 1rem 👍

PR #22, opened 2024-10-23
CMS-440: data model and migrations

diego-oxd:
### Jira Ticket

CMS-440
https://bcparksdigital.atlassian.net/browse/CMS-440

### Description
- added data model
- added sequelize-cli to manage DB migrations
- need to run `npx sequelize-cli db:migrate` to sync changes

Model definitions are in the models folder, sequelize-cli creates a basic template of the model and then more details have to be added manually (FK associations, allownull, default values, etc). Same thing with the migrations, sequelize-cli will create a basic template and additional features are added manually. 

`models/index.js` initializes the models and creates the associations so that they are ready to be used by the express API and admin.js. 

An additional `package.json` has to be added to the `migrations` folder. Otherwise running the migrations just throws an error.


duncan-oxd:
Cool! I guess we can get rid of the example models in db/models and the couple of places that are using the "DbRow" test model. If you don't get a chance, I can probably do that tomorrow. Might as well do it in this same branch.

duncan-oxd:
I just pushed a small commit to add a "migrate" npm script, so we can type `npm run migrate` and it will use the version of sequelize-cli installed in package.json 👍

diego-oxd:
> Cool! I guess we can get rid of the example models in db/models and the couple of places that are using the "DbRow" test model. If you don't get a chance, I can probably do that tomorrow. Might as well do it in this same branch.

Cool, I'll do it early tomorrow.

backend/models/index.js
duncan-oxd, line 1: I don't have any experience with Sequelize myself, so I don't know if this is true, but I thought we could do the associations in the model files without needing this extra step. (Unless this makes it easier?)

Instead of an `associate` method in the model file, we would just import the associated models in the file and add a line like 

```js
ParkFeature.belongsTo(Park, { foreignKey: "parkId" });
```

Right after the class definition. Would that be easier?

backend/config/config.js
duncan-oxd, line 1: I think the sequelize connection stuff is already set up in `/db/connection.js` - can we use that? Or is that only gonna be used by Express?

duncan-oxd (changes requested):
Got a few questions and suggestions; forgive me for being clueless because I haven't used Sequelize before :P 

Too bad the models and the db schema aren't as tightly coupled as with Django. I think Sequelize is popular for being easy, but maybe it's "easy but not simple" 🤔 

It's not a requirement for this project if you think we should check out something else like Prisma, but it looks like it should do the trick. Let me know if you have concerns!

backend/models/index.js
diego-oxd: That was my original plan too. But when you run the CLI command to create the migrations it creates the models using this syntax, and then you have to define the associations like this. 
To run something like 
`ParkFeature.belongsTo(Park, { foreignKey: "parkId" });` you'd still need to instantiate every model first like we're doing in index.js. So, I thought might as well just stay in the framework of what `sequelize-cli` suggests. 
We can talk about it tomorrow.

backend/config/config.js
diego-oxd: When you run the migrate CLI command, it always expects this file to exist. I also tried to just reuse the existing sequelize connection. But seems like it requires this file. 

I'm also not familiar with sequelize, but if it's possible, it'd be great.

duncan-oxd: Just pushed some changes to organize this 👍

backend/models/user.js
```diff
@@ -0,0 +1,29 @@
+import { Model, DataTypes } from "sequelize";
+
+export default (sequelize) => {
+  class User extends Model {
+    /**
+     * Helper method for defining associations.
+     * This method is not a part of Sequelize lifecycle.
+     * The `models/index` file will call this method automatically.
+     * @param models
+     */
```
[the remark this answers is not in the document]
duncan-oxd: I just realized these files are probably auto-generated... If the comment format disagrees with eslint, we can add an exception to the rules to ignore these.

duncan-oxd (approved):

PR #23, opened 2024-10-23
CMS-427: Build "Preview" page elements

duncan-oxd:
### Jira Ticket

CMS-427

### Description

The last page of elements/styles to build out (I think). The next step will be to set up the API and then turn these static HTML elements into React components when the API is ready.

### DIFF
--- a/frontend/src/router/pages/ReviewChanges.scss
+++ b/frontend/src/router/pages/ReviewChanges.scss
@@ -1,0 +1,47 @@
+.page.review-changes {
+  .sub-area {
+    margin-bottom: var(--layout-margin-large);
+    padding-block: var(--layout-padding-medium);
+
+    & > h2 {
+      margin-bottom: var(--layout-margin-xlarge);
+      display: flex;
+      gap: var(--layout-margin-medium);
+
+      .sub-area-icon {
+        height: 1.25em;
+      }
+    }
+  }
+
+  .date-range {
+    display: flex;
+    justify-content: space-between;
+    width: fit-content;
+
+    .date {
+      min-width: 6em;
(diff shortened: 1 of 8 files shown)


ayumi-oxd (approved):
LGTM 👍 A button "Back to {PARK NAME} dates" is needed on the top of the preview page but we can add it later.

frontend/src/router/pages/ReviewChanges.scss
```diff
@@ -0,0 +1,47 @@
...
+  // spacing between date range values
+  td:not(:last-child) .date-range {
+    margin-right: 1em;
+  }
+
+  .notes {
+    .note-metadata {
+      color: var(--typography-color-secondary);
+      filter: brightness(1.5); // muted font color, lighter than color-secondary
+    }
```
ayumi-oxd, line 41: Very cool, I didn't know this 😮

duncan-oxd: Ideally there would be a color in the design tokens for this, but the "primary" and "secondary" font colors look pretty much the same... Another thing I'll mention to Amanda when we go through things :P I think a custom shade of grey is probably fine to have for this.

SCSS has more color functions, but you'd need the colors to be SCSS variables to use them!
https://sass-lang.com/documentation/modules/color/

duncan-oxd:
I'll add that "<- back to..." nav button on all of the pages in this branch, since it's the last ticket under CMS-422 👍

duncan-oxd:
Just pushed a commit to add the "Back" links to the pages that needed it. I'll merge this and then go over the current state of everything with Amanda 👍

PR #24, opened 2024-10-25
CMS-439: add read endpoints

diego-oxd:
### Jira Ticket

CMS-


backend/routes/api/parks.js
```diff
@@ -0,0 +1,133 @@
...
+      ],
+    });
+
+    if (!park) {
+      throw new Error(`Park not found: ${orcs}`);
+    }
+
+    const ouput = park.get({ plain: true });
+
+    res.json({ ouput });
```
duncan-oxd, line 129:
```suggestion
    const output = park.get({ plain: true });

    res.json({ output });
```
😄

backend/package.json
duncan-oxd, line 1: We can remove this change from the branch

backend/seeders/package.json
duncan-oxd, line 1: Looks like something's up with the line endings in your git/editor settings?

PR #28, opened 2024-11-01
CMS-451: Build EditAndReview table components

duncan-oxd:
### Jira Ticket

CMS-451

### Description

This branch hooks up the /parks API route (and adds a custom `useApi` hook to query it).
It replaces the hardcoded Bootstrap table with a react component that displays the data and links to the "park details page" route.

**It adds an environment variable to the frontend for `VITE_API_BASE_URL` so you'll need to update your local .env** - see .env.example or grab the new one from 1Password


frontend/src/hooks/useApi.js
duncan-oxd, line 1: useEffect goes crazy if you pass in an object as a dependency, so stringified it here and had to add an eslint-disable line. Please let me know if there's a better pattern for doing that. (`params` is an object with unknown properties)

And also, in React strict mode, useEffect fires twice when the page loads so I added `isFetching` to prevent that. Again, let me know if there's a better way to do that!

frontend/src/router/pages/EditAndReview.jsx
```diff
@@ -39,57 +104,37 @@ function EditAndReview() {
...
-                  Approved
-                </span>
-              </td>
-              <td>&nbsp;</td>
-            </tr>
-          </tbody>
-        </table>
-      </div>
+      {loading ? (
+        <p>Loading...</p>
```
duncan-oxd, line 134: I'll ask Amanda if we have specs for a loading status, otherwise I'll just pop in a typical spinner thing.

duncan-oxd:
Forgot we also need to sort the table by the park name :\ The rest is still ready to look at, though

frontend/src/router/pages/EditAndReview.jsx
ayumi-oxd: Just FYI, we used this animated progress bar: https://digitalspace.github.io/bcparks-bootstrap-theme on the BC Parks Find a Park page https://bcparks.ca/find-a-park, but of course, we don't have to use the same one both on public and internal sites.

frontend/src/components/EditAndReviewTable.jsx
```diff
@@ -0,0 +1,109 @@
...
+  // navigate to park details page
+  function navigateToPark() {
+    navigate(getParkLink());
+  }
+
+  return (
+    <tr onClick={navigateToPark} role="button">
+      <th scope="row">
+        <Link to={getParkLink()}>{park.name}</Link>
+      </th>
```
ayumi-oxd, line 23: A park name seems to be `#2D2D2D` in the mock.

frontend/src/router/pages/EditAndReview.jsx
```diff
@@ -29,67 +97,42 @@ function EditAndReview() {
...
+          >
+            <option value="">All</option>
 
-        <div className="col-lg-3 col-md-6 col-12">
-          <label htmlFor="bundle" className="form-label">
-            Bundle
-          </label>
-
-          <select id="bundle" className="form-select">
-            <option value=""></option>
```
ayumi-oxd, line 43: After the keycloak implementation is done (has it?), we could add a function to check a user's role (e.g. Park Operator/Admin) and display "Bundle" and "Has reservation" in the table if a user has the admin role. We did some thing similar in the SP. 

https://github.com/bcgov/bcparks-staff-portal/blob/main/frontend-legacy/src/auth/PrivateElement.js
https://github.com/bcgov/bcparks-staff-portal/blob/main/frontend-legacy/src/components/page/appDashboard/AppDashboard.js

ayumi-oxd (approved):
Left some notes but everything LGTM 👏

frontend/src/hooks/useApi.js
```diff
@@ -0,0 +1,40 @@
...
+      try {
+        isFetching.current = true;
+        const response = await axiosInstance.get(endpoint, { params });
+
+        setData(response.data);
+      } catch (err) {
+        setError(err);
+      } finally {
+        setLoading(false);
+      }
```
ayumi-oxd, line 30: Do we need to reset `isFetching.current` to false in the `finally`?

frontend/src/components/EditAndReviewTable.jsx
duncan-oxd: I wrapped it in a link tag and then the color changed (and added an underline). I think it _should_ have some indication that you can click on it, but I'll check with design and see if we should change the color or style

frontend/src/router/pages/EditAndReview.jsx
```diff
@@ -1,7 +1,73 @@
...
+    data?.map((park) => ({
+      ...park,
+      status: `TODO: Active ${park.id}`,
+    })) || [];
+
+  // unique values for filter dropdowns
+  const statusValues =
+    parks
+      .map((park) => park.status)
+      .filter((value, index, self) => self.indexOf(value) === index) || [];
```
diego-oxd, line 40: Can we try using a Set? I think it'd be a bit more efficient.

frontend/src/router/pages/EditAndReview.jsx
```diff
@@ -1,7 +1,73 @@
...
+    if (
+      nameFilter &&
+      !park.name.toLocaleLowerCase().includes(nameFilter.toLocaleLowerCase())
+    ) {
+      return false;
+    }
+
+    if (statusFilter && park.status !== statusFilter) {
+      return false;
+    }
```
diego-oxd, line 52: From the wireframes it seems the user can select more than one status to filter.

diego-oxd:
What do you think of a pattern like this. It kind of looks cleaner and it's pretty popular in React. Especially when you have a hook for fetching data like the one you added here. 

```
const MyComponent = () => {
  const [data, loading, error] = useFetchAPI('api/parks');

  if (loading) {
    return <div>Loading...</div>;
  }

  if (error) {
    return <div>Error: {error.message}</div>;
  }

  return (
    <div>
      {data.map((item) => (
        <div key={item.id}>{item.name}</div>
      ))}
    </div>
  );
};
```

frontend/src/router/pages/EditAndReview.jsx
duncan-oxd: I chatted with Amanda today and she said that bar should be fine for our purposes. I'll add it in a separate branch and replace this "Loading..." text 👍

frontend/src/components/EditAndReviewTable.jsx
duncan-oxd: I talked to Amanda today, we're thinking the link hover color in the bootstrap theme is incorrect (according to the Design Tokens page) - I'll make a small ticket to override it and fix it site-wide.
CMS-468 - I'll grab it in the next sprint

duncan-oxd: We also decided to add an icon to indicate that the row is clickable, so I might remove the Link tag. I'll make that update along with the loading animation.

duncan-oxd:
I definitely prefer that early return pattern to putting big indents and "&&" in the template. I'll clean it up a bit and push that with some other cosmetic updates I discussed with Amanda today 👍

frontend/src/router/pages/EditAndReview.jsx
duncan-oxd: Updated this branch with a `LoadingBar` component after discussing with Amanda 👍

frontend/src/router/pages/EditAndReview.jsx
duncan-oxd: The users table / auth stuff will be separate; I think we need to work out a way to pull a bunch of data from Strapi first. Right now you have to log in with your IDIR, but nothing is checking roles or any other permissions.

frontend/src/router/pages/EditAndReview.jsx
duncan-oxd: 💀 I hooked up a crude dropdown/checkbox group with native bootstrap stuff. Ayumi showed me a third party component we can probably use if we need more functionality than that.

frontend/src/router/pages/EditAndReview.jsx
duncan-oxd: I just changed this to a hardcoded list of all of the status values in the end. That previous code was more useful for when the values were unknown

diego-oxd (approved):
Looks great. 

In some cases, we could use `useMemo` to avoid recalculating values or arrays. But for now, I don't this should be an issue and it'd be super easy to add in the future. 

Feel free to merge if it's ready.

duncan-oxd:
I forgot about useMemo 😊 I'm used to computed properties all being memoized automatically 😄  We can tackle that in a quick ticket later on once we get all the pages working together

PR #29, opened 2024-11-05
CMS-372: Pass link error status to validator

ayumi-oxd:
### Jira Ticket
CMS-372

### Description
- Pass link error status to validator
- Copied over from https://github.com/bcgov/bcparks.ca/pull/1511


duncan-oxd (approved):
I don't have experience with this validator API myself but the code looks sensible 👍

PR #30, opened 2024-11-07
CMS-452: Build Park Details components

duncan-oxd:
### Jira Ticket

CMS-452

### Description

This branch builds out components for the Park Details page, and hooks it up to data from the API.

I added a useDate hook to let us parse and format dates. We can add other date functions in there as needed.
(I thought about automatically parsing all dates inside the API hook, but that may not be necessary for this MVP)

It also adds a StatusBadge component and replaces the placeholder content on the "Edit and Review" table landing page. It has a hardcoded list of possible status strings.

## Two things from the ticket that aren't done:

> Global variable for reviewer 

I think this sill need to be done after we import data from Strapi (or elsewhere). Let me know if there's anything I should do in the meantime.

> Displayed on expansion of accordion 
> - Summary of dates 
> - Notes  

Do we have an endpoint that provides this data already? Should I add one? I'd prefer to make it a separate ticket if it still needs to be built.

### DIFF
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -1,4 +1,5 @@
 import { Router } from "express";
+import _ from "lodash";
 import { Park, Season, FeatureType, Feature } from "../../models/index.js";
 import asyncHandler from "express-async-handler";
 
@@ -104,9 +105,17 @@ router.get(
       throw error;
     }
 
-    const output = park.toJSON();
+    const parkJson = park.toJSON();
 
-    res.json(output);
+    const subAreas = _.mapValues(
+      // group seasons by feature type
+      _.groupBy(parkJson.seasons, (s) => s.featureType.name),
+
+      // sort by year
+      (group) => _.orderBy(group, ["operatingYear"], ["desc"]),
+    );
+
+    res.json({ ...parkJson, subAreas });
   }),
 );
 
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -115,6 +115,9 @@ router.get(
       (group) => _.orderBy(group, ["operatingYear"], ["desc"]),
     );
 
+    // remove unused season key
+    delete parkJson.seasons;
+
     res.json({ ...parkJson, subAreas });
   }),
 );
--- a/frontend/src/components/ParkDetailsSeason.jsx
+++ b/frontend/src/components/ParkDetailsSeason.jsx
@@ -7,12 +7,24 @@ import PropTypes from "prop-types";
 import { useNavigate, useParams } from "react-router-dom";
 import StatusBadge from "@/components/StatusBadge";
 import useDate from "@/hooks/useDate";
+import { useApiGet } from "@/hooks/useApi";
+import LoadingBar from "@/components/LoadingBar";
+import SeasonDates from "@/components/ParkDetailsSeasonDates";
 
 export default function ParkSeason({ season }) {
   const { parkId } = useParams();
   const navigate = useNavigate();
 
   const [expanded, setExpanded] = useState(false);
+
+  const {
+    data: datesData,
+    loading,
+    error,
+    fetchData,
+  } = useApiGet(`/seasons/${season.id}`, {
+    instant: false,
+  });
 
   // Returns a chevron icon based on the expansion state
   function getExpandIcon() {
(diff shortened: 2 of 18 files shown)


backend/routes/api/parks.js
```diff
@@ -104,9 +105,17 @@ router.get(
...
-    res.json(output);
+    const subAreas = _.mapValues(
+      // group seasons by feature type
+      _.groupBy(parkJson.seasons, (s) => s.featureType.name),
+
+      // sort by year
+      (group) => _.orderBy(group, ["operatingYear"], ["desc"]),
+    );
+
+    res.json({ ...parkJson, subAreas });
```
duncan-oxd, line 118: Temporary change in here to format the data differently. If this looks okay, we can make it permanent

frontend/src/components/ParkDetailsSeason.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+        {expandableContent()}
+      </div>
+
+      <div className="controls">
+        <button onClick={navigateToEdit} className="btn btn-text-primary">
+          <FontAwesomeIcon className="append-content me-2" icon={faPen} />
+          <span>Edit</span>
+        </button>
+
```
ayumi-oxd, line 123: When I opened the season, the controls element height slightly changed. Just wanted to know if it's intentional.
Screenshot 2024-11-08 at 3 44 03 PM

frontend/src/components/ParkDetailsSubArea.jsx
```diff
@@ -0,0 +1,29 @@
...
+import walkInCamping from "@/assets/icons/walk-in-camping.svg";
+import PropTypes from "prop-types";
+
+export default function SubArea({ title, data }) {
+  return (
+    <section className="sub-area">
+      <h2>
+        <img src={walkInCamping} className="sub-area-icon" /> {title}
+      </h2>
+
```
ayumi-oxd, line 12: Would be great if we could use API data to display icons.

frontend/src/components/StatusBadge.jsx
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  const classes = classNames(["badge", "rounded-pill", colorClass]);
+
+  return <span className={classes}>{label}</span>;
+}
+
+StatusBadge.propTypes = {
+  status: PropTypes.string,
+};
```
ayumi-oxd, line 34: Isn't the park status supposed to be required? or could it be without status?

ayumi-oxd (approved):
Looks great! Just some nitpicky questions.

frontend/src/components/StatusBadge.jsx
duncan-oxd: I asked for clarification on Slack because I'm actually not sure if there are any cases where it could be blank

frontend/src/components/ParkDetailsSubArea.jsx
duncan-oxd: Agreed 👍 I think those values should come over with other metadata when we sync our data with Strapi, but right now we don't have a field for it.

frontend/src/components/ParkDetailsSeason.jsx
duncan-oxd: The contents in the left panel are technically dynamic height, but I made the right panel have a taller minimum height to make it much less likely to have an obvious change in height

backend/routes/api/parks.js
diego-oxd: If we're returning the seasons grouped by their featureType. Then, we probably can get rid of the regular list of seasons to reduce the payload.

frontend/src/components/ParkDetailsSeasonDates.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+function DateTypeRow({ dateType }) {
+  console.log("dateType", dateType);
+
+  return (
+    <tr>
+      <td>{dateType.dateType.name} dates</td>
+      <td>
+        <DateRange start={dateType.startDate} end={dateType.endDate} />
+      </td>
```
diego-oxd, line 14: A single dateType and season can have more than one date. 
Screenshot 2024-11-14 at 1 40 12 PM

frontend/src/components/ParkDetailsSeasonDates.jsx
```diff
@@ -0,0 +1,57 @@
...
+// "Dateable" features: Campsite groupings, etc.
+function CampGroundFeature({ feature }) {
+  return (
+    <div className="feature">
+      <h5>{feature.name}</h5>
+
+      <table className="table table-striped sub-area-dates mb-0">
+        <tbody>
+          {feature?.dateable?.currentSeasonDates?.map((dateType) => (
+            <DateTypeRow key={dateType.id} dateType={dateType} />
```
diego-oxd, line 28: The data from the API is not grouped by dateType. It would have a list of dates for the current season with both reservation and operation dates included there. So, a filter is probably needed for each type.

backend/routes/api/parks.js
duncan-oxd: Just pushed an update to delete "seasons" from the structure ✅

diego-oxd:
Left a couple of comments mostly related to the data structures. Lmk if I misunderstood something. 

Also regarding the `useDate`, I don't have super strong opinion about it, but I'd lean towards using a pure function that just takes the date and returns the formatted string. 

I think a hook is useful if we had dates that are dynamically changing often. However, the dates don't change at all. They only change in the main form view, and it that case we need to do some custom manipulation because we're using that 3rd party component. Again, not a strong opinion if you want to stick with the hook, but I think it'd just be easier and cleaner to use a pure function.

Everything else looks great!

frontend/src/components/ParkDetailsSeasonDates.jsx
duncan-oxd: Woah, I definitely didn't finish this file 😆 I'll clean it up.

frontend/src/components/ParkDetailsSeasonDates.jsx
duncan-oxd: If the placeholder data I put into the seasons.js endpoint is still accurate, then I'll group it on the frontend 👍 

I added a few extra dates to simulate a campsite grouping with 3 sets of dates so I'll update that.

duncan-oxd: Pushed some updates to that file, now it will render multiple DateRange components in a single DateTypeRow if there are more than one

duncan-oxd:
After both of our PRs merge, I'll do another small PR to follow up and combine our date formatting stuff into a set of helper functions, and I'll update the whole app 👍

diego-oxd (approved):

PR #31, opened 2024-11-08
CMS-327: main form view (edit/submit season dates)

diego-oxd:
### Jira Ticket

CMS-327

### Description
Backend changes
- Added models for keeping track of changes (SeasonChangeLog and DateChangeLog).
  - each DateChangeLog is associated to one SeasonChangeLog (this would be like a transaction or form submission)
  - added respective associations for those models (User, Season)
- added `readyToPublish` field to the season. This will let staff decide which of the approved seasons get published to the API. 
- updated `/seasons/:seasonId` endpoint to match the structure needed on the frontend. 
- added new endpoint `/seasons/:seasonId/save/` to save changes and create change logs. 

Frontend changes
- added flash message confirmation
- added new slide toggle component
- added new `refetchData` method to the `useGetAPI `hook to refresh the page
- added new method to send post requests.
- added different functions to format dates
- the main form view has 2 main state components:
   - season
   - dates (decoupled from the main season object so that when we updaåte a date, we don't have to reconstruct the whole object every time)


backend/routes/api/seasons.js
```diff
@@ -7,11 +7,20 @@ import {
...
+  SeasonChangeLog,
+  DateChangeLog,
+  User,
 } from "../../models/index.js";
 import asyncHandler from "express-async-handler";
+import { Op } from "sequelize";
 
 const router = Router();
 
+function getNewStatusForSeason(season, user) {
```
duncan-oxd, line 20: Can we add a comment for what this method is supposed to do (or what it will eventually do)?

frontend/src/components/ConfirmationDialog.jsx
duncan-oxd, line 1: If there's another branch for implementing the confirmation dialog, and if it's unused in this branch, then let's take it out of the branch and stash it somewhere to use later

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -0,0 +1,94 @@
...
+}
+
+.button-primary {
+  background-color: #1a365d;
+  color: white;
+}
+
+.button-primary:hover {
+  background-color: #152a4a;
+}
```
duncan-oxd, line 94: We shouldn't need any of these styles since we've already got Bootstrap providing them with "btn"
https://getbootstrap.com/docs/5.3/components/buttons/

frontend/src/components/SlideToggle.jsx
duncan-oxd, line 1: I was just using the built-in bootstrap toggle switch for this; see the one on the Review/Preview page. 
I'm fine using whichever - I don't like toggle switches either way 😆 If we keep this one, we might eventually need to add the "On"/"Off" label to it, but let's not worry about that now.

frontend/src/components/ConfirmationDialog.scss
duncan-oxd, line 1: For any scss files we're adding, let's add a classname to the root component (.modal-dialog or .confirmation-dialog in this case) and then wrap the whole file in a selector, so that the styles like `.close-button` don't interfere with other scopes

```sass
.confirmation-disloag {
  //... file contents
}
```

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -0,0 +1,94 @@
...
+  justify-content: center;
+  z-index: 1000;
+}
+
+.modal-content {
+  background-color: white !important;
+  border-radius: 8px;
+  width: 100%;
+  max-width: 60vw;
+  padding: 24px;
```
duncan-oxd, line 20: We'll need to use the values from the Design Tokens list as much as possible in our styles:
https://www2.gov.bc.ca/gov/content/digital/design-system/foundations/design-tokens/glossary
For example, this line would be `padding: var(--layout-padding-large);`
(turning the dots into dashes: "layout.padding.large" -> "--layout-padding-large")

Check out the existing .scss files in the frontend for some examples. Let me know if you want a hand and we can divide and conquer these files tomorrow

frontend/src/hooks/useApi.js
duncan-oxd, line 1: I added a method to allow manually fetching (or re-fetching, I suppose) data in my PR:
https://github.com/bcgov/bcparks-staff-portal/blob/92bf411cdc87d5cbef86c428ad4cbd6d6b5b5eef/frontend/src/hooks/useApi.js

We could use that, and also build out the post method into a `useApiPost` hook with the same patterns as the GET one.

ayumi-oxd:
@diego-oxd It's not mandatory but we have some recommended rules for a git commit message.
No need to fix your old commits, but this would be helpful in the future 🙂: https://docs.google.com/document/d/1RbmW8h00d4MiHjjQNFg9FOEgZk8T6WUdnNj1hIr8ES8/edit?usp=sharing

PR #35, opened 2024-11-18
CMS-468: Override theme link hover color

duncan-oxd:
### Jira Ticket

CMS-468

### Description

A small change to update the link/text hover color and make it align with the `surface.color.primary.hover` value from the design tokens page. The bootstrap theme had it using a very bright blue that looked out of place.

I also got a warning about Sass syntax so I re-ordered some lines in DateRange.scss while I was at it. It's an unrelated change but I was fixing small style issues in this branch anyway...

### DIFF
--- a/frontend/src/components/DateRange.scss
+++ b/frontend/src/components/DateRange.scss
@@ -4,10 +4,11 @@
   width: fit-content;
 
   .date {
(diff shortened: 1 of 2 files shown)


ayumi-oxd (approved):
LGTM!

PR #37, opened 2024-11-19
CMS-205: Update Bootstrap to v5

ayumi-oxd:
### Jira Ticket
CMS-205

### Description
- Remove bootstrap
- Add bcparks-bootstrap-theme
- Change some Bootstrap class names: https://getbootstrap.com/docs/5.0/migration/
- Change some headings
- Delete yarn.lock file


frontend-legacy/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@bcgov/bc-sans": "^2.1.0",
-    "@bcgov/bootstrap-theme": "https://github.com/bcgov/bootstrap-theme/releases/download/v1.1.1/bcgov-bootstrap-theme-1.1.1.tgz",
+    "@bcgov/bootstrap-v5-theme": "^1.0.1",
```
duncan-oxd, line 7: Is this the same package? looks like 1.0.1 is the latest on NPM but it was previously using 1.1.1?

frontend-legacy/yarn.lock
duncan-oxd, line 1: There shouldn't be both a package.lock and a yarn.lock here 🤨 Which tool are we using for the project? The readme suggests we use `npm install` so we should delete this yarn lockfile to avoid confusion if that's the case.

ayumi-oxd: Good catch! One time we changed npm to yarn but apparently, we forgot to delete it ...

duncan-oxd (approved):
The code changes look good to me! I don't know the site well enough to spot any differences with the bootstrap theme, but I'm sure QA will let us know if they find any

PR #39, opened 2024-11-27
CMS-322: Add form validation hook

duncan-oxd:
### Jira Ticket

CMS-322

### Description

This branch adds validation functions for the "SubmitDates" page ("notes" field and the date ranges). The logic turned out to be complicated to implement, even if the rules themselves aren't too bad. 

In our meetings last week, we decided it would be unnecessary hassle to reformat the data into a format to work with React Hook Form, and then write our own custom validators for the data structure (which is dynamic, and deeply-nested). The result is that this whole thing is custom-made, which makes it potentially fragile and error prone! But the benefit is that we can have our data structured the way we want, and run validations without needing to change it and then change it back.

I struggled with the state hooks, updating asynchronously when a field is changed, and then instantly trying to validate. In order to validate the newly-changed values, I added a callback function to `updateDateRange`, which feels sketchy but seems to be a common practice. Let me know if there's a cleaner way to do this.

There's a lot to look at here, so I'm hoping for a review from @diego-oxd especially, but @ayumi-oxd if you get some time even to skim it, I'd welcome any feedback!

## Update from Dec 11:

One `@TODO` items remains right now: 

**Scroll the first error into view:** React Hook Form does this. If the validation errors are outside the viewport when you click the button at the bottom, there should be some indication of why your submission didn't work. Either scroll the errors into view, or show something next to the button to indicate validation errors.

This isn't part of the current requirements, so I'll merge without it but I'll mention to Amanda and see if it's something we want to do in a future ticket.

### DIFF
--- a/frontend/src/hooks/useValidation.js
+++ b/frontend/src/hooks/useValidation.js
@@ -147,7 +147,7 @@ export default function useValidation(dates, notes, season) {
     // Check if the start date is before the end date
     if (startDate > endDate) {
       return addError(
-        endDateId,
+        dateRangeId,
         "Enter an end date that comes after the start date",
       );
     }
@@ -160,7 +160,7 @@ export default function useValidation(dates, notes, season) {
       endDate.getFullYear() !== operatingYear
     ) {
       // startDate =< endDate check happens first, so the end date will never fail this check
-      return addError(dateRangeId, `Enter dates for ${operatingYear} only`);
+      return addError(startDateId, `Enter dates for ${operatingYear} only`);
     }
 
     return true;
@@ -233,7 +233,7 @@ export default function useValidation(dates, notes, season) {
 
     if (daysBetween < 1) {
       return addError(
-        dateableFeature,
+        dateableId,
         "Reservation end date must be one or more days before the operating end date.",
       );
     }
--- a/frontend/src/hooks/useValidation.js
+++ b/frontend/src/hooks/useValidation.js
@@ -1,6 +1,6 @@
 import { useState } from "react";
-import { omit, mapValues, minBy, maxBy } from "lodash";
-import { differenceInCalendarDays } from "date-fns";
+import { omit, mapValues, minBy, maxBy, orderBy } from "lodash";
+import { differenceInCalendarDays, parseISO, isBefore, max } from "date-fns";
 
 // Validation functions for the SubmitForm component
 export default function useValidation(dates, notes, season) {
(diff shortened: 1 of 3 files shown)


frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -160,32 +165,43 @@ function SubmitDates() {
...
+    callback = null,
+  ) {
+    const newValue = value ? value.toISOString() : null;
+
+    setDates((prevDates) => {
+      const updatedDates = cloneDeep(prevDates);
+
+      // Update the date value and mark the date range as changed
+      lodashSet(updatedDates, [dateableId, dateType, index, key], newValue);
+      lodashSet(updatedDates, [dateableId, dateType, index, "changed"], true);
```
duncan-oxd, line 186: I was having so much trouble with stale state, I swapped this code out for Lodash - I don't think it made a difference but it's more readable anyway 🤷

frontend/src/hooks/useValidation.js
```diff
@@ -0,0 +1,298 @@
...
+        const maxDate = maxDateRange?.endDate
+          ? new Date(maxDateRange.endDate)
+          : null;
+
+        return { minDate, maxDate };
+      }),
+    );
+
+    return mapped;
+  }
```
diego-oxd, line 58: I know you mentioned this is on TODO to account for gaps for non-continuous date ranges, but I think we should include it on this PR. Probably just need to check for each reservation date range one at a time and make sure that it's covered.

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -160,32 +166,43 @@ function SubmitDates() {
...
+
+      if (callback) {
+        callback(updatedDates);
+      }
```
diego-oxd, line 191: I think it'd be better to use `useEffect` with the dates dependency. That's pretty much a listener that will be triggered when anything in the dates object changes.

diego-oxd:
Screenshot 2024-11-30 at 9 28 57 AM

I get this error when both of reservation fields are empty. I believe this should be allowed to add operation dates only.

diego-oxd:
It looks good. I agree that points 1, 4, and 5 can be left for later. 
But I think we should include 2 and 3 here. 
We might need to check in with Amanda. But it does feel confusing when you see highlighted errors on the fields that you are not currently editing. Also, you might see the same error multiple times when you have more than one dateRange for operation or reservation. 

I don't it if makes sense to have field-level errors and feature-level errors. Or maybe just highlight the error on the field being edited? Just my thoughts, but we should probably check with Amanda.

duncan-oxd:
> I get this error when both of reservation fields are empty. I believe this should be allowed to add operation dates only.

I've just messaged Amanda to see if we want to delay the validation until after the user clicks submit the first time. You'll need to enter 4 dates before the form can possibly be "valid" so there are a lot of errors that might confuse things as you go.

frontend/src/router/pages/SubmitDates.jsx
duncan-oxd: I tried a few different approaches but I kept running into trouble with stale state. In this case I wanted to set it up so you could specify the callback (or have no callback at all), but it still feels clunky... I will remember it for refactoring in the future, and hopefully I'll learn a better pattern in the meantime, haha.

frontend/src/hooks/useValidation.js
duncan-oxd: I just pushed an update that does a more robust check on the date ranges: 
1. sort the dates chronologically
2. combine any overlaps
3. _then_ make sure all reservation ranges are within an operating date range

frontend/src/hooks/useValidation.js
```diff
@@ -0,0 +1,341 @@
...
+      // startDate =< endDate check happens first, so the end date will never fail this check
+      return addError(startDateId, `Enter dates for ${operatingYear} only`);
+    }
+
+    const dateType = dateRange.dateType.name;
+    const { dateableId } = dateRange;
+
+    // Check if the reservation dates are all within the operating dates
+    const dateableFeature = datesObj[dateableId];
+    const reservationDatesValid = validateReservationDates(dateableFeature);
```
diego-oxd, line 180: Would it make more sense to call this function inside of validateFeatureDates only once?
Right now, validateFeatureDates calls validateDateRange for every dateRange in the feature and `validateReservationDates` would be called for every dateRange even though is basically validating the same thing for the whole feature.

frontend/src/hooks/useValidation.js
```diff
@@ -0,0 +1,341 @@
...
+    });
+
+    // If the changed dateRange is invalid, don't validate anything else
+    if (!rangeValid) return false;
+
+    // If the changed dateRange is valid, validate the whole campsite grouping feature
+    // to resolve any inter-dependent date range validations.
+
+    // @TODO: This unnecessarily validates the changed dateRange twice;
+    // Consider refactoring to validate once, but only show relevant errors.
```
diego-oxd, line 321: It could be easier to just call validateFeatureDates for every change to avoid calling validateDateRange twice. Realistically, each feature would have at most like 4 dates? So, it wouldn't be a big performance issue.
But, I'm not totally against the duplicate calls either if we return false as soon as we find an error in the dateRange being edited.

frontend/src/hooks/useValidation.js
duncan-oxd: There's a fair bit of unnecessary repeat validating happening right now... I'd like to maybe make it just validate the entire form (or the entire campsite grouping section, at least) once on every change, since so many rules depend on each other. It would un-tangle the code a fair bit, and then we could easily debounce it if we want. I'm gonna wait to hear from UX about the updates to the error displays. If we only need to display an error at the feature level, then it makes sense to just check errors at that level too. (We wouldn't need to track the ID of the field that changed)

I agree with you, though! I'll check with Amanda and see if that's something we'll be doing in this ticket, or a bit later.

frontend/src/hooks/useValidation.js
duncan-oxd: Just updated it to validate the whole feature at once on change, instead of starting with just the one date range

duncan-oxd:
Just rebased and pushed some changes to address all of the earlier TODO items, except one mentioned in the description above. If this looks good, I think we can merge it and then tidy up any errors that arise after we pull in the Strapi data

diego-oxd (approved):

duncan-oxd:
The requirements for this ticket  have changed again; I'm going to make a small tweak and just push it, and then I'll tackle the rest of the changes in a separate branch.

PR #40, opened 2024-12-02
CMS-454: script to sync data with strapi

diego-oxd:
### Jira Ticket

CMS-454

### Description
There are 3 models in our DB that have a direct relationship with their respective Strapi models. For those 3 models (`Park`, `Feature`, `FeatureType`), I added a `strapiId` field to make synchronization easier. 

For the models `Season` and `DateRange`. There is a script that should only run one time. 
Seasons don't have a direct mapping to Strapi, but we can create Seasons based `operatingYear`, `featureType`, and `park` from existing dates in Strapi.
DateRanges are created based on `park-operation-sub-area-date`, but from the single  Strapi model we create 1 or 2 local objects (for different date types). 

DateTypes don't have anything similar on Strapi. For now, we are just creating `Operation` and `Reservation` types. 

TODO: 
- still need to create campgrounds, but this is something that will need to be done manually and wouldn't be part of the script.


backend/strapi-sync/schema.js
```diff
@@ -0,0 +1,222 @@
...
+  {
+    model: "protected-area",
+    endpoint: "/protected-areas",
+    fields: [
+      "orcs",
+      "protectedAreaName",
+      "type",
+      "typeCode",
+      "parkFacilities", // FK (don't use it?)
+      // "parkNames", // FK
```
duncan-oxd, line 99: Should these comments be updated?

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -130,6 +130,7 @@ function SubmitDates() {
               dateableId: feature.dateable.id,
               inputType: "text",
               changed: false,
+              validationError: null,
```
duncan-oxd, line 133: Heads up: this variable will be removed in the validation branch anyway

backend/strapi-sync/sync.js
```diff
@@ -0,0 +1,404 @@
...
+
+  try {
+    let currentUrl = url;
+
+    if (queryParams) {
+      currentUrl = `${currentUrl}?${queryParams}`;
+    }
+
+    const response = await get(currentUrl);
+    const data = await response.json();
```
duncan-oxd, line 42: Looks like this is for `fetch` but axios response data is just a plain object, isn't it?

backend/strapi-sync/sync.js
```diff
@@ -0,0 +1,404 @@
...
+    const pagination = data.meta.pagination;
+
+    items.push(...data.data);
+
+    let page = 1;
+
+    while (page < pagination.pageCount) {
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      page += 1;
+
```
duncan-oxd, line 52: We could use a `for` loop to save having to manually increment `page`:
```suggestion
    for (let page = 1; page < pagination.pageCount; page++) {
      await new Promise((resolve) => setTimeout(resolve, 1000));

```

I'm guessing that 1s pause is to avoid overloading the API? Let's throw some comments in to say why it's doing that

backend/strapi-sync/sync.js
```diff
@@ -0,0 +1,404 @@
...
+      let paginatedUrl = currentUrl;
+
+      if (queryParams) {
+        paginatedUrl = `${paginatedUrl}&pagination[page]=${page}`;
+      } else {
+        paginatedUrl = `${paginatedUrl}?pagination[page]=${page}`;
+      }
```
duncan-oxd, line 59: Since there are a lot of places where we're working with query parameters, we could consider using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and possibly [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) to build the request URLs. (I figure we could append stuff to the query string without needing checks like this)

backend/strapi-sync/sync.js
```diff
@@ -0,0 +1,404 @@
...
+
+          queryParams.push(param);
+        }
+
+        qpString = queryParams.join("&");
+      }
+
+      const temp = getData(currentUrl, qpString);
+
+      item.items = temp;
```
duncan-oxd, line 146: Can we just combine this onto one line?
```suggestion
      item.items = getData(currentUrl, qpString);
```

backend/strapi-sync/sync.js
```diff
@@ -0,0 +1,404 @@
...
+
+  return dbItem;
+}
+
+export async function syncParks(parkData) {
+  const items = parkData.items;
+
+  items.forEach(async (item) => {
+    await createOrUpdatePark(item);
+  });
```
duncan-oxd, line 179: An async function implicitly returns a Promise, so `forEach` won't actually wait for the calls to `createOrUpdatePark` to finish here and it will run them all at the same time.  If we really want to `await` and run each call one-by-one, we can change this to a `for..of` loop. If we _do_ want to run them all in parallel, we could take out the `async`/`await`, but in that case we should use `Promise.all` so we can catch exceptions.

Looks like this has a decent explanation if you're curious: https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971

backend/strapi-sync/sync.js
```diff
@@ -0,0 +1,404 @@
...
+    };
+
+    dbItem = await createModel(Feature, data);
+  }
+
+  return dbItem;
+}
+
+export async function syncFeatures(featureData) {
+  const items = featureData.items.slice(400, 700);
```
duncan-oxd, line 273: What are the 400 and 700 from? Add a comment to explain what's going on there

duncan-oxd (commented):
Lookin' great! I put in some suggestions to consider about changing a few things to more ✨ idiomatic javascript ✨ but, only if it makes things easier.

And you already know what I'm gonna say 😆 Let's clean up the commented-out lines in sync.js and add some code comments to everything so the script will be easy to follow at a glance.

backend/strapi-sync/create-campgrounds.js
duncan-oxd, line 1: Looks like this isn't used anywhere..?

backend/strapi-sync/sync.js
```diff
@@ -70,6 +72,10 @@ export async function getData(url, queryParams) {
   }
 }
 
+/**
+ * Fetches data for all the models of interest asynchrously
```
duncan-oxd, line 76:
```suggestion
 * Fetches data for all the models of interest asynchronously
```

backend/strapi-sync/sync.js
```diff
@@ -227,11 +239,14 @@ export async function createDateTypes() {
...
 
-  data.forEach(async (item) => {
-    await createModel(DateType, item);
-  });
+  await Promise.all(data.map((item) => createModel(DateType, item)));
 }
 
+/**
+ * For the feature in strapi, create a new feature or update its corresponding existing feature
+ * @param {Object} item For the feature in strapi, create a new feature or update its corresponding existing feature
```
duncan-oxd, line 247: Looks like a copy/paste error on this line

backend/strapi-sync/sync.js
```diff
@@ -291,6 +316,7 @@ export async function createDatesAndSeasons(datesData) {
 
   await Promise.all(
     items.map(async (item) => {
+      // if there is no subAreaId, we can't create a date range the date is orphaned
```
duncan-oxd, line 319:
```suggestion
      // if there is no subAreaId, we can't create a date range and the date is orphaned
```
?

duncan-oxd (approved):
Looks good! I haven't run it locally yet, so be sure to include instructions on how/when to import Strapi data when you update the backend Readme in CMS-459 👍

PR #42, opened 2024-12-05
CMS-433: Define breakpoint mixins

duncan-oxd:
### Jira Ticket

CMS-433

### Description

A quick ticket from a while ago when we were building the layouts; this adds some sass mixins we can use for mobile-first responsive layouts, so we don't have to memorize the breakpoint widths.

The regular Bootstrap ones aren't available here because we're only using the Bootstrap theme and not pulling in the base Bootstrap library with all the individual sass files.

I updated the Vite config so they'll be automatically available in every scss file.


frontend/src/styles/breakpoints.scss
duncan-oxd, line 1: I swear Sass lets you define mixins from a loop, but this project is using `sass-embedded` and it doesn't seem to work here 🤷 So I just manually defined all 5

diego-oxd (approved):

PR #43, opened 2024-12-05
CMS-453: Update status change logic and add not-ready flag

diego-oxd:
### Jira Ticket

CMS-453

### Description
- `/seasons` endpoint now returns `readyToPublish` field.
- `/parks` endpoint only returns parks with subAreas that have dates
- status is changed to "requested" after user saves any change. 
- Added not-ready flag component to seasons in which `readyToPublish` is false.
- fixed text for mark approved button
- fixed bug in which `readToPublish` was not being properly set.

### DIFF
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -54,7 +54,7 @@ router.get(
         {
           model: Season,
           as: "seasons",
-          attributes: ["id", "status"],
+          attributes: ["id", "status", "readyToPublish"],
         },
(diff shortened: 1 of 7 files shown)


duncan-oxd:
Looks good - I remember Amanda mentioned some requirements that aren't in the ticket so I've left a comment to confirm if we still want to show the flag on the homepage as well (In the parks list).

duncan-oxd (approved):

PR #47, opened 2024-12-09
CMS-566: Fix expansion panel bug to show all features

diego-oxd:
### Jira Ticket

CMS-566

### Description
Fixed expansion panel bug. Previously was only showing features that belong to campgrounds. Now, it should show every other type of feature as well.


duncan-oxd (approved):
Nice, I bet this will reveal some oversights in the validation functions 🫠 but it's good that it's fixed!

PR #48, opened 2024-12-10
CMS-517: add modal windows and flash messages

diego-oxd:
### ### Jira Ticket

CMS-517

### Description
Both modal windows and flash messages have their own hook. You can use their how to show the element on screen and pass params. It's basically just a way to avoid duplication and send different prop values to the components on demand. 

I also created a hook for the components in which we need to warn the user when they leave the page with unsaved changes. 
- navigating away by clicking a button or a link will show the custom dialog 
- browser-level actions like closing the tab or refreshing would only show the default browser warning

You can check in the ticket where are flash messages and modals supposed to appear. 
The only one that is not included in this PR, because it relates to submitting for review, and that's not part of this version is:
> Submit with missing dates? When there are empty date fields and they clicked submit for review
> Continue editing - return to form
> Submit - Submitted for review and return to park details

### DIFF
--- a/frontend/src/components/ConfirmationDialog.scss
+++ b/frontend/src/components/ConfirmationDialog.scss
@@ -11,34 +11,24 @@
   z-index: 100;
 }
 
-.confirmation-dialog-outer {
-  display: grid;
+.confirmation-dialog-container {
   grid-template-columns: 1% 99%;
   border-radius: var(--layout-border-radius-large);
   box-shadow:
     0 4px 6px -1px rgba(0, 0, 0, 0.1),
     0 2px 4px -1px rgba(0, 0, 0, 0.06);
 
+  background-color: white;
+  padding: var(--layout-padding-large);
+  border-radius: var(--layout-border-radius-large);
+  border-left: 5px solid var(--support-border-color-warning);
+
   max-width: 80%;
   width: 100%;
+
   @include from-md {
     max-width: 50%;
   }
-}
-
-.confirmation-dialog-container {
-  background-color: white;
-  padding: var(--layout-padding-large);
-  border-top-right-radius: var(--layout-border-radius-large);
-  border-bottom-right-radius: var(--layout-border-radius-large);
-}
-
-.confirmation-dialog-left-warning-column {
-  background-color: var(--support-border-color-warning);
-  display: inline-block;
-  padding: 10px;
-  border-top-left-radius: var(--layout-border-radius-large);
-  border-bottom-left-radius: var(--layout-border-radius-large);
 }
 
 .confirmation-dialog-title {
@@ -69,5 +59,4 @@
   border: none;
   color: var(--icons-color-secondary);
   font-size: 32px;
-  // float: right;
 }
--- a/frontend/src/components/ConfirmationDialog.scss
+++ b/frontend/src/components/ConfirmationDialog.scss
@@ -1,0 +1,61 @@
+.confirmation-dialog-overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0.5);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  z-index: 100;
+}
+
+.confirmation-dialog-container {
+  background-color: white;
+  padding: 24px;
+  border-radius: 8px;
+  box-shadow:
+    0 4px 6px -1px rgba(0, 0, 0, 0.1),
+    0 2px 4px -1px rgba(0, 0, 0, 0.06);
+  max-width: 50%;
+  width: 100%;
+
+  @media (max-width: 767px) {
+    max-width: 80%;
+  }
+}
+
+.confirmation-dialog-title {
+  font-size: 18px;
+  font-weight: bold;
+  margin-bottom: 16px;
+}
+
+.confirmation-dialog-message {
+  margin-bottom: 24px;
+}
+
+.confirmation-dialog-actions {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.confirmation-dialog-cancel-btn {
+  background-color: transparent;
+  color: #6b7280;
+  padding: 8px 16px;
+  border-radius: 4px;
+  margin-right: 8px;
+  cursor: pointer;
+  transition: background-color 0.3s;
+}
+
+.confirmation-dialog-cancel-btn:hover {
+  background-color: #f3f4f6;
+}
+
+.confirmation-dialog-confirm-btn:hover {
+  background-color: #2563eb;
+}
--- a/frontend/src/components/ConfirmationDialog.scss
+++ b/frontend/src/components/ConfirmationDialog.scss
@@ -2,8 +2,8 @@
   position: fixed;
   top: 0;
   left: 0;
-  width: 100%;
-  height: 100%;
+  bottom: 0;
+  right: 0;
   background-color: rgba(0, 0, 0, 0.5);
   display: flex;
   justify-content: center;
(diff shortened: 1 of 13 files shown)


frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -68,18 +92,39 @@ function SubmitDates() {
...
+    if (["under review", "approved", "published"].includes(season.status)) {
+      openConfirmation(
+        "Move back to draft?",
+        "The dates will be moved back to draft and need to be submitted again to be reviewed.",
+        () => {
+          saveChanges(savingDraft);
+        },
+        "If dates have already been published, they will not be updated until new dates are submitted, approved, and published.",
+      );
+    } else {
```
duncan-oxd, line 125: I edited some dates and clicked the "Save draft" button and this confirmation popped up, but it saved at the same time (the /save request showed in the network tab, and the success flash message came up before I clicked cancel or confirm)

diego-oxd: I changed the logic for showing the modal everywhere else but here. Anyway, just pushed the fix.

frontend/src/components/ConfirmationDialog.scss
duncan-oxd, line 1: The Figma design has a big X button in the corner, let's use "fa xmark" for that https://fontawesome.com/icons/xmark?f=classic&s=regular and have it do the same thing as the cancel button.

The designs also have a subtle yellow border on the left side (the "support-border-color-warning" yellow)

https://fontawesome.com/icons/xmark?f=classic&s=regular

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -0,0 +1,56 @@
+.confirmation-dialog-overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
```
duncan-oxd, line 6: You can define top/bottom/left/right all as 0 to get this same effect, but it's a little safer in case the parent element does something weird with padding

```suggestion
  top: 0;
  left: 0;
  bottom: 0;
  right: 0;
```

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -0,0 +1,56 @@
...
+  border-radius: var(--layout-border-radius-large);
+  box-shadow:
+    0 4px 6px -1px rgba(0, 0, 0, 0.1),
+    0 2px 4px -1px rgba(0, 0, 0, 0.06);
+  max-width: 50%;
+  width: 100%;
+
+  @media (max-width: 767px) {
+    max-width: 80%;
+  }
```
duncan-oxd, line 26: Let's keep the styles all "mobile first" and use media queries to progressively enhance styles as the viewport grows instead of shrinks. We can use the media queries I added recently:

```suggestion
  max-width: 80%;
  width: 100%;

  @include from-md {
    max-width: 50%;
  }
```

The idea is that the baseline styles will work for all screens including the smallest, and the media queries control what happens on larger screens. (the `from-md` mixin is the same as `min-width: 768px`).

This pattern is very common, and mostly we just want to avoid mixing the two patterns in one project because that gets really chaotic 😄

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -0,0 +1,56 @@
...
+  color: var(--surface-color-primary-default);
+  padding: 8px 16px;
+  border-radius: var(--layout-border-radius-default);
+  margin-right: var(--layout-margin-small);
+  transition: background-color 0.3s;
+}
+
+.confirmation-dialog-confirm-btn:hover {
+  background-color: var(--surface-color-primary-default);
+}
```
duncan-oxd, line 56: Looks like these styles aren't used?

frontend/src/router/pages/PreviewChanges.jsx
```diff
@@ -82,7 +109,10 @@ function PreviewChanges() {
     });
     setNotes("");
     fetchData();
-    setShowFlash(true);
+    openFlashMessage(
+      "Dates saved as draft",
+      `${data?.park.name} ${data?.featureType.name} ${data?.operatingYear} season details saved`,
```
duncan-oxd, line 114: I suppose `data` will always be fully populated by the time this is showing anyway, but FYI the `?.` optional chaining operator doesn't "short circuit" the whole chain so in cases like this you'd have to put them at every `.` in the chain like this:

```suggestion
      `${data?.park?.name} ${data?.featureType?.name} ${data?.operatingYear} season details saved`,
```

Otherwise if `park` is null or undefined, it will try to read ".name" from it and throw a TypeError.

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -68,18 +92,40 @@ function SubmitDates() {
 
     const response = await sendData(payload);
 
+    if (savingDraft) {
+      setNotes("");
+      fetchData();
+      openFlashMessage(
+        "Dates saved as draft",
+        `${season?.park.name} ${season?.featureType.name} ${season?.operatingYear} season details saved`,
```
duncan-oxd, line 100: Same note as above with the optional chaining. We could use some kind of conditional check to make sure `season` is set, or otherwise show some other string... but I don't think that case can ever happen. Maybe it's safe to take out the `?`s all together?

duncan-oxd (changes requested):
Cool! Some minor notes, mostly style tweaks. Looks like a familiar pattern to what we were doing with Vuetify previously 👍

frontend/src/router/pages/PreviewChanges.jsx
diego-oxd: I guess I didn't think this was necessary because, park and featureType should always be there. If park or featureType are not part of the data nothing else should work either and we should expect to see an error.

frontend/src/router/pages/SubmitDates.jsx
diego-oxd: Same here, season is just the data for the whole page. So, it can happen that season is null. But it could never happen season is not null and park is null. 
I'm not a huge fan of using `?` anyway. Maybe at one point, it'd be better to just do all the checks at the beginning and not worry about optional chaining operators.

duncan-oxd: Even with the optional chaining, you'd end up with a sentence with missing words. In most cases in the template, it's easier and safer to just wrap it in something like `{season && (<blah blah>)}` so it won't even render until the data is loaded.

frontend/src/router/pages/PreviewChanges.jsx
duncan-oxd: If data is always set in this context, then we don't need any `?`s. All or nothin', I guess 👍

diego-oxd: It could happen that data is null because it depends on the hook to fetch the data. But, once we know that `data` is there, park and featureType should always be there. So, I think it makes sense to just have the `?` after `data`. This is consistent with how it's done in other places of the app.  
I can just add all the ? after every object if you like, but I don't think it would make a difference.
I found a weird state that happens only rare occasions in which loading is false, error is null, and data is null. So, I don't think we can get rid of all of the `?` until we figure that out.

duncan-oxd: I just mean that the question mark before "park" wouldn't help if `data` was null because of the way optional chaining works. You can try something like this in the console:

image

So as long as we guard that chunk of the template to not render until `data` is populated, we can ditch the question marks. Or we can add them all the way through the chain, but having it only part way would cause problems because it doesn't "short circuit" and it keeps checking all he way down the chain.

frontend/src/router/pages/SubmitDates.jsx
duncan-oxd: All good! I learned something new about optional chaining today 😅 🥴

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -2,33 +2,55 @@
...
   align-items: center;
   z-index: 100;
 }
 
-.confirmation-dialog-container {
-  background-color: white;
-  padding: var(--layout-padding-large);
+.confirmation-dialog-outer {
+  display: grid;
+  grid-template-columns: 1% 99%;
```
duncan-oxd, line 16: You can actually accomplish that yellow edge effect just with something like
```css
  border-left: 1rem solid var(--support-border-color-warning);
```

This approach works too, but the percentages will make its size change on different screens, so let's hardcode it to a constant value (looks like 5px in Figma)

```suggestion
  grid-template-columns: 5px 1fr;
```

frontend/src/components/ConfirmationDialog.scss
```diff
@@ -42,15 +64,10 @@
...
-  margin-right: var(--layout-margin-small);
-  transition: background-color 0.3s;
-}
-
-.confirmation-dialog-confirm-btn:hover {
-  background-color: var(--surface-color-primary-default);
+  border: none;
+  color: var(--icons-color-secondary);
+  font-size: 32px;
+  // float: right;
```
duncan-oxd, line 72:
```suggestion
```

duncan-oxd (approved):
That was a tricky rebase,  haha. If anything got messed up, we'll fix it after everything is merged together today.

duncan-oxd:
I made a backup of the pre-rebase branch, just in case I made a mistake resolving the merge conflicts 👍

PR #50, opened 2024-12-12
CMS-332: Update validation UI

duncan-oxd:
### Jira Ticket

CMS-332

### Description

More requirements emerged for the validation ticket, so I've implemented them here:

1. Show a red "*" next to the notes field when it's required (when the dates have been approved or submitted already)
2. Add a ⚠️ icon to the range- and feature-level error messages, and highlight the related fields in red
3. Add the missing calendar icon back to the DatePicker inputs
4. Turn the feature-level error message into a bootstrap alert-style box

### DIFF
--- a/frontend/src/hooks/useValidation.js
+++ b/frontend/src/hooks/useValidation.js
@@ -131,6 +131,11 @@ export default function useValidation(dates, notes, season) {
     endDateId,
     dateRangeId,
   }) {
+    // Skip validation for empty ranges
+    if (!start && !end) {
+      return true;
+    }
+
     // Both dates are required if one is set
     if (!start) {
       return addError(startDateId, "Enter a start date");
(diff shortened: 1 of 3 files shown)


diego-oxd:
Looks good. The only issue I found is that you still get an error on the first field (startDate) even if the error is on the endDate. 
Screenshot 2024-12-18 at 3 32 44 PM

duncan-oxd:
> Looks good. The only issue I found is that you still get an error on the first field (startDate) even if the error is on the endDate. 

😂 Oh wow, I thought the end date could never be invalid because of some earlier checks, but I didn't consider that it could be too far in the future. I split the check up so it will catch both cases and highlight the correct field now.

I also just pushed the update we talked about, which skips validation on ranges where both dates are blank. We'll need to "handle" those blank ranges in another ticket.

diego-oxd (approved):

diego-oxd:
Before you merge, can you please try to merge the one with the modal windows. I spent a couple hours trying to rebase and solve conflicts in that one, and this one shouldn't have much of a difference with alpha on the `SubmitDates` page.

duncan-oxd:
💪 Can do, I'll sort out any rebasing that needs to happen tomorrow

PR #52, opened 2024-12-16
CMS-583: create campgrounds

diego-oxd:
### Jira Ticket

CMS-583

### Description
Please refer to the ticket to see the logic behind the grouping. 
There are 2 scripts that create campgrounds and assign features to them.
1. For campgrounds that only have one feature, the campground will keep the name of the single feature and the feature will be renamed to "All Sites". This can be optionally displayed on the FE. 
2. For campgrounds that have multiple features, the campground will have it's own general name, and features will have their name that only reflects their specific part of the campground (e.g. "Alice Lake Campground Sites 1-20" -- > "Sites 1-20").

### DIFF
--- a/backend/strapi-sync/create-multiple-item-campgrounds.js
+++ b/backend/strapi-sync/create-multiple-item-campgrounds.js
@@ -1,4 +1,4 @@
-import { Campground, Feature } from "../models/index.js";
+import { Campground, Feature, Park } from "../models/index.js";
 
 import { getItemByAttributes, createModel } from "./utils.js";
 
@@ -412,21 +412,27 @@ async function updateFeature(item, campgroundId) {
 }
 
 async function createCampground(item) {
+  const park = await getItemByAttributes(Park, {
+    strapiId: item.parkId,
+  });
+
   // create campground with FK to Park
   const data = {
     name: item.campgroundName,
-    parkId: item.parkId,
+    parkId: park.id,
   };
 
   const campground = await createModel(Campground, data);
 
   await Promise.all(
-    item.items.map((feature) => updateFeature(feature, campground.id)),
+    item.items.map(
+      async (feature) => await updateFeature(feature, campground.id),
+    ),
   );
 }
 
 async function createCampgrounds(items) {
-  await Promise.all(items.map((item) => createCampground(item)));
+  await Promise.all(items.map(async (item) => await createCampground(item)));
 }
 
 createCampgrounds(campgrounds);
--- a/backend/strapi-sync/create-multiple-item-campgrounds.js
+++ b/backend/strapi-sync/create-multiple-item-campgrounds.js
@@ -1,4 +1,8 @@
-const items = [
+import { Campground, Feature } from "../models/index.js";
+
+import { getItemByAttributes, createModel } from "./utils.js";
+
+const campgrounds = [
   {
     parkId: 6,
     campgroundName: "Alice Lake Campground",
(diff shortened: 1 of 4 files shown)


backend/strapi-sync/sync.js
```diff
@@ -258,7 +258,13 @@ export async function createOrUpdateFeature(item) {
   let dbItem = await getItemByAttributes(Feature, { strapiId: item.id });
 
   if (dbItem) {
-    dbItem.name = item.attributes.parkSubArea;
+    // if dbItems has campgroundId, don't update the name
+    // there is a script to create campgrounds and assign features to them
+    // this script will rename the feature to fit within the campground
+    if (!dbItem.campgroundId) {
+      dbItem.name = item.attributes.parkSubArea;
+    }
```
duncan-oxd, line 266: I noticed while pulling in the Section/Mgmt Area details for the CSV, we need to make sure we're running `dbItem.save()` after making changes, don't we?

backend/strapi-sync/create-multiple-item-campgrounds.js
```diff
@@ -0,0 +1,438 @@
+import { Campground, Feature, Park } from "../models/index.js";
+
+import { getItemByAttributes, createModel } from "./utils.js";
+
+const campgrounds = [
```
duncan-oxd, line 5: Is this hardcoded list customized? Or can it come from Strapi somehow?

backend/strapi-sync/create-single-item-campgrounds.js
```diff
@@ -0,0 +1,1846 @@
...
+    strapiId: item.items[0].featureStrapiId,
+  });
+
+  feature.campgroundId = campground.id;
+  feature.name = "All sites";
+  await feature.save();
+}
+
+async function createSingleItemsCampgrounds(items) {
+  await Promise.all(items.map(async (item) => await createCampground(item)));
```
duncan-oxd, line 1843: Technically we don't need the async/await in the callback, because we just want it to return a Promise anyway. It won't make a difference besides making the line a little shorter if you want to change it

duncan-oxd:
> For campgrounds that have multiple features, the campground will have it's own general name, and features will have their name that only reflects their specific part of the campground (e.g. "Alice Lake Campground Sites 1-20" -- > "Sites 1-20").

That's a good idea, that's more comprehensive than I thought it would be 👍

backend/strapi-sync/create-multiple-item-campgrounds.js
diego-oxd: It has to be hard-coded because it was created based on "made up" arbitrary rules. The data was manually reviewed to create this list. It helps to keep track the strapiId of the features we are changing. 
Also, I think Brett made some edits, so this will need to be updated.

backend/strapi-sync/sync.js
diego-oxd: Yes, you're right. I could have sworn I fixed that somewhere. But, anyway just pushed the fix.

backend/strapi-sync/create-multiple-item-campgrounds.js
duncan-oxd: Got it 👍 And we had discussed before, if campsite groupings change in the future, hopefully we can update our db first and be the source of truth, rather than changing Strapi first and throwing things out of whack in our db.

backend/strapi-sync/create-multiple-item-campgrounds.js
```diff
@@ -425,14 +425,12 @@ async function createCampground(item) {
   const campground = await createModel(Campground, data);
 
   await Promise.all(
-    item.items.map(
-      async (feature) => await updateFeature(feature, campground.id),
-    ),
+    item.items.map(async (feature) => updateFeature(feature, campground.id)),
```
duncan-oxd, line 428: You won't need the `async` on these lines either. The `async` keyword makes a function return a promise, but `updateFeature` already returns a promise so we can remove here.

PR #54, opened 2024-12-18
CMS-348: Add CSV export feature

duncan-oxd:
### Jira Ticket

CMS-348

### Description

The CSVs require some info we weren't pulling from strapi, and we don't need that info anywhere else so I didn't want to two new FK tables and a junction table just for that. I tried a few approaches and this was the cleanest/simplest one:
- Added a jsonb field to the Park model called "managementAreas" which stores an array like this:

```json
[
  {
    "mgmtArea": {
      "strapiId": 123,
      "name": "North Okanagan"
    },
    "section": {
      "strapiId": 57,
      "name": "Okanagan"
    }
  }
]
```

The details get fetched in the normal syncData script, and the CSV export just prints out the names from that array.

In particular, I'd like someone to look at the Sequelize query in the Export endpoint, because was guessing at some of the sub-query/join stuff. I've been testing it since I imported the real Strapi data and it looks okay, but let me know if you notice anything strange.

Other stuff:
- Updated the eslint config to stop showing errors in the seeders/migrations/.adminJs folders with auto-generated files.
- Refactored the Express routes slightly so they're all sub-routes of /api/ and we don't need to add /api/ and the base path to every new route we define.
- Tweaked the fluid width of the flash message so it looks better with the long message on the Export page

### DIFF
--- a/backend/eslint.config.mjs
+++ b/backend/eslint.config.mjs
@@ -30,4 +30,26 @@ export default [
       "new-cap": "off",
     },
   },
+
+  // Limit linting in seeders/migrations directories
+  {
+    files: ["migrations/**/*.js", "seeders/**/*.js"],
+
+    languageOptions: {
+      globals: { ...globals.node },
+
+      parserOptions: {
+        sourceType: "script", // Auto-generated files are CJS modules
+      },
+    },
+
+    rules: {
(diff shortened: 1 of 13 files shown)


duncan-oxd:
Rebased and resolved conflicts. The requirements changed again this morning so I'll push another update and then merge it 🫠

PR #56, opened 2024-12-19
CMS-603: Fix dates timezone issue. Use UTC dates regardless of timezone.

diego-oxd:
### Jira Ticket

CMS-603

### Description
We will basically use UTC everywhere. This is how they are stored in the DB and how they come back from the API. We're making sure that we when format them, the UTC date is displayed and it's not being converted to PST time.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -18,6 +18,7 @@
         "axios": "^1.7.5",
         "classnames": "^2.5.1",
         "date-fns": "^4.1.0",
+        "date-fns-tz": "^3.2.0",
         "lodash": "^4.17.21",
         "oidc-client-ts": "^3.0.1",
         "prop-types": "^15.8.1",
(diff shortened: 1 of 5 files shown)


duncan-oxd (changes requested):
Works for me! All UTC or all BC time on both sides avoids confusion. 
The validation functions do some date parsing in a few places; can you update them to use the `normalizeToUTCDate` function too? Just search the project for `new Date(` and there are maybe 2 files with a few spots to update. There's also a couple of spots in the validation hook that use `parseISO`, if that makes a difference 🤔 

Was it the formatISO or the datepicker that was mixing things up? It's weird that we'd pick a local date and then it would get sent to the DB as UTC time in the first place, but I didn't even notice.

diego-oxd:
> Works for me! All UTC or all BC time on both sides avoids confusion. The validation functions do some date parsing in a few places; can you update them to use the `normalizeToUTCDate` function too? Just search the project for `new Date(` and there are maybe 2 files with a few spots to update. There's also a couple of spots in the validation hook that use `parseISO`, if that makes a difference 🤔
> 
> Was it the formatISO or the datepicker that was mixing things up? It's weird that we'd pick a local date and then it would get sent to the DB as UTC time in the first place, but I didn't even notice.

Sure, I'll update in a bit. 
It was basically always using UTC. But whenever you display a date, JS will automatically convert the date (even if it's UTC) to you local timezone.

duncan-oxd:
The frontend is creating and editing dates locally, so it was kind of confusing that it would take "Jan 1, at 00:00 Pacific" and send it to the DB to be stored as "Jan 1, 00:00 UTC" (Which is really Dec 31 at 16:00 Pacific).

I put something in the backlog for when we're refactoring: we can parse all of the dates one time when they're received from the backend, and then we don't have to worry about what's a string and what's adjusted, etc. 

I was wrong about the validation; I realized it was mostly passing around dates that are already parsed and so adjusting them again would make them off by 8 hours from where we need them. Really I just had to change `getFullYear` to `getFullUTCYear` and the rest was already working. I've got a commit ready that I'll push up to fix that in a moment.

PR #57, opened 2024-12-19
CMS-591: Add ChangeLogs (notes) component to details page

duncan-oxd:
### Jira Ticket

CMS-591

### Description

We had to add the change log notes to the details page, in the expansion panels. I turned the notes list into a component and popped it in there, and made some minor style tweaks to make it match the latest Figma styles.

### DIFF
--- a/frontend/src/components/ChangeLogs.jsx
+++ b/frontend/src/components/ChangeLogs.jsx
@@ -1,0 +1,29 @@
+import PropTypes from "prop-types";
+import { formatTimestamp } from "@/lib/utils";
+
+// Displays season date changelogs (aka "notes")
+export default function ChangeLogs({ changeLogs = [] }) {
+  return (
+    <>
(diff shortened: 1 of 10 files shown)


duncan-oxd:
...and pushed an update to rename the component so it doesn't have the same name as the prop 😅

diego-oxd:
Looks good. The only thing is that the date and name under the note should be a different colour than the note. It looks fine on the SubmitDates page and on the PreviewPage, but on the expansion panels they're both the same colour. 

Screenshot 2024-12-19 at 7 25 36 PM

duncan-oxd:
> the date and name under the note should be a different colour than the note. It looks fine on the SubmitDates page and on the PreviewPage, but on the expansion panels they're both the same colour.

Ah, good catch. I just need to move the styles from the pages into the ChangeLogsList component 👍

PR #59, opened 2024-12-20
CMS-566: fix issue in which some seasons didn't have any dates at all

diego-oxd:
### Jira Ticket

CMS-566

### Description
This changes the one-time-data-import script so that seasons and dates will only be created for active features. Some inactive features have missing dates or dates that are also inactive. 
Given that the API considers which features are active and which features have dates, we don't want to create "seasons" for features that wouldn't be returned by the API anyway. 

Note that this requires the database to be recreated and all the syncing process to be done from scratch. Just follow steps 1 to 7 in the readme. 

The `required: false`  in the query is to make the query a left outer join instead of an inner join. This was causing 500 errors given that in some cases dateable was being returned as `null` even though the feature had a dateable.

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -97,6 +97,7 @@ router.get(
     const parkFeatures = await Feature.findAll({
       where: {
         parkId: season.park.id,
+        active: true,
         featureTypeId: season.featureType.id,
       },
       attributes: ["id", "name", "hasReservations", "active"],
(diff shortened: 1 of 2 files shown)


duncan-oxd:
I discovered that `required` thing when I was doing the query for the CSV too :D it was kind of hard to get my head around it in a nested code structure compared to just writing a query, but it makes sense now.

duncan-oxd (approved):

PR #60, opened 2024-12-21
CMS-592: Update date format for "Previous" ranges

duncan-oxd:
### Jira Ticket

CMS-592

### Description

Added the year to the date formats, only for the "Previous: " date range string.


ayumi-oxd (approved):
LGTM!

PR #65, opened 2024-12-24
CMS-572: Add FeatureIcon component

duncan-oxd:
### Jira Ticket

CMS-572

### Description

Added a component to resolve the `featureType.name` string from Strapi to a local svg path to display the icon. It involved a big map but I think it's fine. The dev server will pre-import all of the files, but in production I _think_ it will just populate the img src so the icons will be loaded on demand.

I also renamed some classnames and variables for clarity.


diego-oxd:
Looks good. 

Note that for wilderness camping, the icon path we get from Strapi is `wilderness-camping.svg` not just `wilderness-camping`. We either have to change it in strapi or change the map here. 
Everything else looks good.

ayumi-oxd:
Change icon name from `wilderness-camping.svg` to `wilderness-camping` on Srapi PROD.

PR #69, opened 2025-01-03
CMS-586: Store Keycloak roles in a global context

duncan-oxd:
### Jira Ticket

CMS-586

### Description

I added an "Access Context" layer between the auth and the rendered templates where we can check specific roles and permissions. Right now it just gets the user's roles out of the Keycloak token and makes them available globally. 
If we want, we can also redirect navigation based on role checks there. And we can store an array of associated parks etc.

I also added a `useAccess` hook to do specific checks throughout the app/template wherever we might need. Right now it's not being used anywhere, but we can do this:

```js
const { checkAccess, ROLES } = useAccess();

if (checkAccess(ROLES.SUPER_ADMIN)) {
  console.log("User is a super admin");

  return <div>Secret button</div>;
}
```

I decided this would be cleaner than manually setting stuff in a store like Redux, which we would have to maintain ourselves. I don't anticipate needing much more than the list of roles, so Redux would be overkill.

I also added the access token to the Authorization header for all requests by default, and protected the routes with the `checkJwt` middleware on the backend. That just checks that the token is valid. The next step will be to check specific roles/permissions to allow requests, but that's a separate ticket in the backlog (CMS-588).


diego-oxd:
Nice. I agree that using context is better and simpler than Redux for this simple scenario.

diego-oxd (approved):

PR #72, opened 2025-01-06
CMS-603: Convert timestamp in change logs to local time when formatting.

diego-oxd:
### Jira Ticket

CMS-603

### Description
Every date and timestamp is saved in UTC. When displaying date in change logs, we need to uses `America/Vancouver` timezone when formatting to display the accurate day.


duncan-oxd (approved):
👍 Please let Manuji know there's this update so she can text CMS-603 again. Looks like that extra parameter will be handy in the future

PR #80, opened 2025-01-09
CMS-599: Add table pagination

duncan-oxd:
### Jira Ticket

CMS-599

### Description

Added some basic pagination to the table on the home page. It sounds like the number of rows per page will be decided once they've had a chance to test it, but for now I've gone with 20.


ayumi-oxd (approved):
LGTM!

PR #83, opened 2025-01-10
CMS-618: Redirect to Park Details on approval

duncan-oxd:
### Jira Ticket

CMS-618

### Description

Instead of staying on the Preview page when you approve dates, now it redirects to the Park Details page.

Most of the changes here were to move the flash message to the Park Details page. I used an "approved" query param and useEffect. I'm not sure I used the hook efficiently so let me know if you think of any improvements.
An improvement for a future version could be to make the flash message global in the layout so we can call it and still change routes.

### DIFF
--- a/frontend/src/router/pages/ParkDetails.jsx
+++ b/frontend/src/router/pages/ParkDetails.jsx
@@ -1,13 +1,12 @@
-import { useParams } from "react-router-dom";
+import { useEffect } from "react";
+import { useParams, useSearchParams } from "react-router-dom";
 import { useApiGet } from "@/hooks/useApi";
 import { useFlashMessage } from "@/hooks/useFlashMessage";
-import { useSearchParams } from "react-router-dom";
 import NavBack from "@/components/NavBack";
 import LoadingBar from "@/components/LoadingBar";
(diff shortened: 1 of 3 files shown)


frontend/src/router/pages/ParkDetails.jsx
```diff
@@ -30,8 +43,42 @@ function ParkDetails() {
...
+    );
+
+    if (!approvedFeature) return;
+
+    openFlashMessage(
+      "Dates approved",
+      `${park.name} ${approvedFeature.featureType.name} ${approvedFeature.operatingYear} season dates marked approved`,
+    );
+  });
+
```
ayumi-oxd, line 72: It doesn't look like it's happening but it might be better to have dependencies to avoid running on every render. Probably `isFlashOpen`?

ayumi-oxd (approved):
Works as expected! Just some minor notes.

frontend/src/router/pages/ParkDetails.jsx
duncan-oxd: I wondered about that... it seems to fire 3 or 4 times regardless, but I suppose it's good to have the dependencies array anyway. I'll push and update to add that.

PR #94, opened 2025-01-28
CMS-647:  sync park-operation-sub-areas across environments

diego-oxd:
### Jira Ticket

CMS-647

### Description

We are trying to sync subareas across environments. The goal is have every env match what prod has. 
On a high level, there are 2 main operations going on for each environment.
1. If a subarea exists in prod, but not in the environment -- > the subarea will be created in the environment
2. If a subarea exists in the env, but not in prod -- > the subarea will be marked as inactive in the environment.

We are also adding a global ID to subareas to make them easier to manage across envs. 

To compare subareas across envs , we use the name and their "parent" park. It was verified using a separate script that no 2 subareas have the same name and the same parent park in any env. 
So, if 2 subareas in different environments have the same name and same "parent" park, we can be sure that the objects refer to the same subarea. 

All comparisons rely on the "parent" park of the subareas. Basically, the pseudocode would be 2 parts. 
```
for envData in envs:
    subareas = envData.subareas
    for subarea in subareas:
        equivalentProdSubarea = findProdSubarea(subarea.park.orcs, subarea.name)
        
        if equivalentProdSubarea:
            subarea.globalId = `{orcs}_{equivalentProdSubarea.id}`
        else:
            subarea.isActive = False    
```

```
for subarea in prodSubareas:
    for envData in envs:
        equivalentEnvSubarea = findEnvSubarea(subarea.park.orcs, subarea.name)
        
        if not equivalentEnvSubarea:
            // use prod subarea as template
            createSubareaInEnv(subarea)
```

The actual code is different but it's the same idea.
There are few data structures that make the comparison easier and more efficient. 

parksWithSubareas: 
```
{
    envName: {
         orcs: {
             id, name, subareas: [ { id, name, subareaTypeId } ]
         }
    }
}
```

```
{
    envName: {
      orcs: {
        id: parkId,
        orcs: orcs,
        name: parkName,
        missingInProd: [subareaName],
        missingHere: [subareaName],
        totalInProd: totalSubareasInProd,
        totalHere: totalSubareasHere
      }
  }
  ```

```
{
    envName: {
      orcs: {
        subareaName: subareaId
      }
    }
  }
  ```

### DIFF
--- a/backend/strapi-sync/cross-env-data-sync.js
+++ b/backend/strapi-sync/cross-env-data-sync.js
@@ -798,6 +798,32 @@ async function getEnvDataToUpdate(
   return result;
 }
 
+async function addGlobalIdToProdSubareas(envs) {
+  // for prod, we don't need any cross env comparisons
+  const prodData = envs.prod.data;
+
+  const subareas = getModelData(prodData, "park-operation-sub-area").items;
+
+  // loop through all subareas in prod
+  for (const subarea of subareas) {
+    const attrs = { ...subarea.attributes };
+
+    const payload = removeKeysFromPayload(attrs);
+
+    const park = getModelDataById(
+      prodData,
+      "protected-area",
+      getParkId(subarea),
+    );
+
+    // add a global ID to the payload - this will be shared across envs
+    // we use its own ID as part of the global ID - prod is the source of truth
+    payload.globalId = `${park.attributes.orcs}_${subarea.id}`;
+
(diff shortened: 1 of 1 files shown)


duncan-oxd (approved):
Looks great! If we only need to run this once and we don't need to run it _on_ the environments themselves then we don't have to merge it, but maybe don't delete the branch for a while, just in case :angel:

PR #95, opened 2025-01-31
CMS-675: Add "Edit" page for Winter fees (old data structure)

duncan-oxd:
### Jira Ticket

CMS-675

### Description

Added a "Submit dates" / "Edit dates" page for Winter fee dates. Very similar to the non-winter seasons, but with a slightly different data structure. I also used a couple of "contexts" to try to make the components more like pure functions. Even if they're still not... The regular season page has them all nested inside the root component, but they're at least separate functions here, thanks to the contexts providing the data to all the children.

This ticket doesn't include any validation, so I have that commented out for now. The "submit" and "save" functions will also be handled in separate tickets, so they just print the data to the console at the moment.

We've discussed simplifying the API endpoint and the data structure even more; so we can be sure to update the frontend at the same time when we do that.

### DIFF
--- a/backend/routes/api/winterFees.js
+++ b/backend/routes/api/winterFees.js
@@ -217,6 +217,9 @@ router.get(
       };
     });
 
+    // Get date types to include in the response
+    const dateTypes = await DateType.findAll();
+
     res.json({
       name: `${currentYear} – ${currentYear + 1}`,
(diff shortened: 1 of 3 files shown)


duncan-oxd:
Closing this as it's been superseded by #98 using the newer backend endpoint and data structure

PR #96, opened 2025-01-31
CMS-467: Add endpoint to get winter season and import winter dates

diego-oxd:
### Jira Ticket

CMS-467

### Description
- added endpoint to get winter season details
- added sh script to run all commands needed to restart your db and import all the data
- added boolean field `hasWinterFeeDates` to `Feature`
- import winter dates from strapi during the `one-time-data-import` command.

### DIFF
--- a/backend/import-all-data.sh
+++ b/backend/import-all-data.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# Exit immediately if any command fails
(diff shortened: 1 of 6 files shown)


duncan-oxd (approved):
Looking good! Now we'll need to update the two frontend pages built so far to work with the new structure 👍

PR #99, opened 2025-02-03
CMS-673: Add "save winter dates" endpoint

duncan-oxd:
### Jira Ticket

CMS-673

### Description

Added the "save" endpoint for winter dates, and the related frontend stuff (flash messages, confirmations, and save functions).

53fbb21 pretty much copies the code from the normal season endpoint, and then I discovered we can use `bulkCreate` with `updateOnDuplicate` to potentially cut down on the number of INSERT/UPDATE queries by a lot, so a35a0c3 updates the endpoint code to use that. Check it out and let me know what you think!

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -1,4 +1,8 @@
 import { Router } from "express";
+import _ from "lodash";
+import asyncHandler from "express-async-handler";
+import { Op } from "sequelize";
+
 import {
   Park,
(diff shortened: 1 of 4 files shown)


duncan-oxd:
I moved the updated code with the bulkCreate pattern to the regular seasons, and then set both "edit" pages to use that endpoint 👍

diego-oxd (approved):

PR #104, opened 2025-02-05
CMS-338: publish tab bug fix that makes build fail

diego-oxd:
### Jira Ticket

CMS-338

### Description
- Fix bug with incorrect import path (the build failed on merging the previous branch)
- Switched back to import data from prod untill we have featureId in every environment in Strapi.


duncan-oxd (approved):
Looks like it needs a rebase, but the new commit looks good 👍

PR #105, opened 2025-02-05
CMS-338: Fix bug with incorrect import path.

diego-oxd:
### Jira Ticket

CMS-338

### Description
- Fix bug with incorrect import path (the build failed on merging the previous branch)
- Switched back to import data from prod untill we have featureId in every environment in Strapi.

### DIFF
--- a/backend/strapi-sync/sync.js
+++ b/backend/strapi-sync/sync.js
@@ -78,7 +78,8 @@ export async function getData(url, queryParams) {
  * @returns {Array} list of all models with thier name, endpoint, and items
  */
 export async function fetchAllModels() {
-  const url = `${process.env.STRAPI_URL}/api`;
+  // const url = `${process.env.STRAPI_URL}/api`;
+  const url = "https://cms.bcparks.ca/api";
 
   const strapiData = [
     {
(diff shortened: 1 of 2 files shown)


diego-oxd:
Sorry, recreating this PR with a different branch because the history in the other branch was too difficult to rebase.

backend/strapi-sync/sync.js
```diff
@@ -599,7 +600,8 @@ async function createTestUser() {
  */
 export async function oneTimeDataImport() {
   // only meant to run once - not needed for regular sync
-  const url = `${process.env.STRAPI_URL}/api`;
+  // const url = `${process.env.STRAPI_URL}/api`;
+  const url = "https://cms.bcparks.ca/api";
```
duncan-oxd, line 604: Can't we just set our env variable to this value? (On my local machine, it already is `STRAPI_URL="https://cms.bcparks.ca"`)

duncan-oxd (approved):

backend/strapi-sync/sync.js
diego-oxd: We would need to set the value of STRAPI_URL to the prod api in every env and it might just make it confusing.

PR #112, opened 2025-02-07
CMS-693: Add user object to request and assign proper user to change log

diego-oxd:
### Jira Ticket

CMS-693

### Description
- Create user in staff portal DB if it doesn't exist using keycloak info
- Add middleware to assign user object to request object
- assign user who sent the request to change log


backend/middleware/users.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Function} next Express next middleware function
+ * @returns {Promise<void>}
+ * @throws {Error} When authentication data is missing or invalid
+ */
+async function usersMiddleware(req, res, next) {
+  try {
+    // We need email from keycloak to identify the user
+    if (!req.auth?.email) {
+      throw new Error("Authentication data missing");
+    }
```
duncan-oxd, line 16: Awesome 👍 Do we need to send the user's email as a query parameter with every request? If so, I suppose we should do that in this branch or none of the API calls will work when we merge. We could update the useApi hook to do it automatically.

backend/index.js
```diff
@@ -58,6 +59,8 @@ app.use("/", homeRoutes); // Health check route(s)
 // API routes
 const apiRouter = express.Router();
 
+apiRouter.use(usersMiddleware);
```
duncan-oxd, line 62: We might only need this for a few endpoints, and it would save the extra query when we're just getting data for the homepage, etc. 

You can add the middleware as a parameter before the final handler in the endpoint files (similar to the checkJwt middleware being added to the whole router on on line 70

```js
router.post(
  "/:seasonId/save/",
  
  // add any middleware here in order, as long as it calls `next()`
  usersMiddleware,

  asyncHandler(async (req, res) => {
```

backend/middleware/users.js
diego-oxd: The auth object is added to the req object by the auth middleware. So, any authenticated request will have the user's keycloak info, which includes the email.

backend/index.js
diego-oxd: This could work, but I kinda wanted to add the middleware in general to everything so that every user who has logged in to our app will have a user created in our DB. Better to keep track who has used it and this middleware can be later updated to get the subset of parks that users have access to.

backend/middleware/users.js
duncan-oxd: Oh! That's a bonus 😆 I didn't notice it was doing that. I guess the email is in the token anyway. Happy friday 🎶

backend/index.js
duncan-oxd: That sounds good to me, it's not an expensive query and it'll be convenient to have whenever we need it. 
In that case, how about we move the `checkJwt` middleware from line 70 and put it just above this line 62 so we have them both being added everywhere with the same syntax 👍

duncan-oxd (approved):

PR #117, opened 2025-02-13
CMS-517: Update modal button labels and fix modal workflow issues

diego-oxd:
### Jira Ticket

CMS-517

### Description
- changing button labels specific to each modal
- changing some titles that were updated in the wireframes
- navigate to park details page when user saves draft
- create new confirmation dialog with text input that will be added to the notes

Copied from my comment on https://bcparksdigital.atlassian.net/browse/CMS-517

- The modal for “Edit public dates on API?” appears twice on the wireframes with different titles “Edit published dates?“. But according to this ticket, the title should be “Edit public dates on API?“.

- I don’t know if there’s a difference between the “Exit without saving?” modal and the “Discard changes?“ modal. Some notes on this.

     - Navigating away by clicking the back button next to “save draft“ or by clicking the BC parks logo would show the “Discard changes?“ modal.

     - Actions like: closing the tab, clicking the back button on the browser, and reloading the page are browser-level events and we cannot guarantee that the custom modal would be displayed for those. So, only the browser alert would be displayed. Like this:

- When the user clicks the “save draft“ button either in the form or the preview page. A flash message should be displayed with “Dates saved as draft“. This ticket mentions that we should navigate back to the park details page. However, the wireframes show the “Dates saved as draft“ flash message on the Form and Preview pages. Since we couldn’t show the flash message and then navigate right away to the park details page, the “Dates saved as draft“ will be displayed in the park details page. (Similar to the “Dates approved” flash message).

- For now, I updated the label of the button “Continue to preview“ to – > “Save and continue to preview“. This change is temporary for this current version and would be changed back when this ticket is implemented. This related to the confusion between preview and review pages. 

https://bcparksdigital.atlassian.net/browse/CMS-577

### DIFF
--- a/frontend/src/hooks/useMissingDatesConfirmation.js
+++ b/frontend/src/hooks/useMissingDatesConfirmation.js
@@ -1,4 +1,3 @@
-import { set } from "lodash";
 import { useState } from "react";
 
 export function useMissingDatesConfirmation() {
@@ -7,7 +6,7 @@ export function useMissingDatesConfirmation() {
   const [inputMessage, setInputMessage] = useState("");
   const [resolvePromise, setResolvePromise] = useState(null);
 
-  function openMisingDatesConfirmation(featureNameList) {
+  function openConfirmation(featureNameList) {
     return new Promise((resolve) => {
       setFeatureNames(featureNameList);
       setResolvePromise(() => resolve); // Store the resolve function
@@ -15,14 +14,14 @@ export function useMissingDatesConfirmation() {
     });
   }
 
-  function handleMissingDatesConfirm() {
+  function handleConfirm() {
     if (resolvePromise) {
       resolvePromise({ confirm: true, confirmationMessage: inputMessage }); // Resolve with true
     }
     setIsOpen(false);
   }
 
-  function handleMissingDatesCancel() {
+  function handleCancel() {
     if (resolvePromise) {
       resolvePromise({ confirm: false, confirmationMessage: inputMessage }); // Resolve with false
     }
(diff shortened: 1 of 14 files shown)


frontend/src/hooks/useMissingDatesConfirmation.js
```diff
@@ -0,0 +1,43 @@
...
+  return {
+    featureNames,
+    setFeatureNames,
+    inputMessage,
+    setInputMessage,
+
+    openMisingDatesConfirmation,
+    handleMissingDatesConfirm,
+    handleMissingDatesCancel,
+    isMissingDatesConfirmationOpen: isOpen,
```
duncan-oxd, line 41: If this whole file is the "missing dates confirmation" context, it would make sense to me to just call these eg "open", "handleConfirm", "handleCancel", etc. and export them with those names, then we can rename them on the other end if we want, when we're using the hook. 

We could also use this pattern, which might make sense here:

```js
// without de-structuring the object
const missingDatesConfirm = useMissingDatesConfirmation();

missingDatesConfirm.open();

{missingDatesConfirm.isOpen} etc.
```

You don't need to change anything here, it's just an idea for the future ✅

frontend/src/router/pages/ParkDetails.jsx
duncan-oxd, line 1: 👍 This is still pretty clean. If we need to add many more types of "saved" messages, we can probably refactor into some kind of multi-purpose "saved flash message" helper or something. (In Vue we'd often put the flash message in the app root so we could fire off a flash message at any time and it wouldn't interfere with navigation. I bet we'll need to do that at some point)

duncan-oxd (changes requested):
The new code makes sense to me! I put in one change request and some random thoughts. Also be sure to run the linter - it looks like there are some stray unused variables and imports hanging around that can be deleted.

frontend/src/router/pages/ParkDetails.jsx
diego-oxd: Yeah, I agree. If the logic for this flash messages changes or becomes more complicated. We can refactor it in a new ticket.

duncan-oxd (approved):

PR #118, opened 2025-02-14
CMS-700: Add strapiFeatureId field to feature and update campgrounds script

diego-oxd:
### Jira Ticket

CMS-700

### Description
Script to create campgrounds need to be updated so that it can be used in all envs.
For this we have to use the new featureId in Strapi instead of the old 'id. For parks, we'll use the orcs because it serves the same purpose (a unique ID that is consistent across envs).

NOTE: even when approved this shouldn't be merged until the new featureId is available in all envs.


duncan-oxd (approved):
Looks good! Since these are unicode files, we _can_ put in a string like "sx̌ʷəx̌ʷnitkʷ" instead of having to escape the characters "sx\u030c\u02b7\u0259x\u030c\u02b7nitk\u02b7" but I'm guessing it did that automatically somewhere and it's probably safer to leave it :D

PR #121, opened 2025-02-19
CMS-706: updatedAt timestamp on Season will be null on creation

diego-oxd:
### Jira Ticket

CMS-706

### Description
- updatedAt timestamp on Season will be null on creation because we want to show that a season has "never" been updated on the DOOT. 
- added a migration to allow null values on that field 
- added a hook on the Season model that sets the value to null on creation


duncan-oxd (approved):
👍 good thinking

PR #126, opened 2025-02-21
CMS-721: Enable admin dev controls based on env variable

diego-oxd:
### Jira Ticket

CMS-721

### Description
Just switched environment variable so that it's something we can use across environments.


duncan-oxd:
Oh, woops, I added the variables to the helm charts and rebased onto alpha, but this was based on main? I'll undo my changes and fix up the helm charts separately...

duncan-oxd (approved):

PR #128, opened 2025-02-21
CMS-641: Separate alpha and main helm charts

duncan-oxd:
### Jira Ticket

CMS-641

### Description

This adds specific helm charts for the alpha environments to separate them from main so they can have their own crunchy DB installations.

I raised the resource limits when we got our quotas increased, and then copied the main charts to create the alpha ones. The only difference in the files is the name of the crunchyDB deployment has "-alpha" at the end.

These changes are already deployed live. I just realized I still hadn't created a PR for this branch!


duncan-oxd:
Actually I'll just merge this, since it's already been live for a while :P

PR #132, opened 2025-02-21
CMS-717: Fix issue where some date types could be missing in the payload

diego-oxd:
### Jira Ticket

CMS-717

### Description
Missing date type on `api/seasons/:id` was causing frontend to crash. 
The issue was happening when a feature has `hasReservations: true` but currently not reservation dates available for that season. This is not an inconsistency and can happen often, especially with new seasons. 
The fix is simply to include both date types in the payload. 

This ticket also includes the fix for CMS-733. 

Using `Season.update()` will update the `updatedAt` timestamp
But using `myseason.save()` will not update it.

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -149,6 +149,20 @@ router.get(
     // we want to get all the date types for this season
     const dateTypes = {};
 
+    // add datetypes from the db to map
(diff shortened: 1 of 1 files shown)


duncan-oxd (approved):
I didn't know that about the timestamp updates... They don't make that very obvious in their documentation!

PR #135, opened 2025-02-22
Cms 721 reset udpated at timestamp

diego-oxd:
### Jira Ticket

CMS-721

### Description
Update Season to not have automatic timestamps and created custom hooks to manage them.
We need:
- updatedAt to be null on creation
- updatedAt to be the current time whenever a season is updated for any reason
- updatedAt to be null again when the season data is reset

The only way to manage this is to disable automatic timestamps and manage manually.

I added hooks that will update the updatedAt flag for every bulkupdate on seasons. 
We need individual season updates to be unaffected by the hook, because we need to set updatedAt to null when the season is reset.

### DIFF
--- a/backend/middleware/adminJs.js
+++ b/backend/middleware/adminJs.js
@@ -57,16 +57,16 @@ const SeasonResource = {
             const seasonId = record.params.id;
 
             // set status to requested for this season
-            await Season.update(
-              {
-                status: "requested",
-              },
-              {
-                where: {
-                  id: seasonId,
-                },
-              },
-            );
+            const season = await Season.findByPk(seasonId);
+
+            season.status = "requested";
+            season.readyToPublish = true;
+            season.updatedAt = null;
+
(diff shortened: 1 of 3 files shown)


backend/middleware/adminJs.js
```diff
@@ -57,16 +57,16 @@ const SeasonResource = {
...
-                },
-              },
-            );
+            const season = await Season.findByPk(seasonId);
+
+            season.status = "requested";
+            season.readyToPublish = true;
+            season.updatedAt = null;
+
+            // updatedAt can only be set to null if we call save(), not with bulkUpdate
```
duncan-oxd, line 66: Is that even true when we create the model with `timestamps: false`?

duncan-oxd (approved):
Looks good, and in some ways it kind of makes some of the queries cleaner.
If we need to do anything really weird, I suppose we can always make a totally custom field just for display like "updatedAtDisplay" but I think it won't get more complicated than this ✅

PR #137, opened 2025-02-25
CMS-711: Add offline_access token scope

duncan-oxd:
### Jira Ticket

CMS-711

### Description

#127 fixed the token refreshing, but the refresh token still expires after 6 hours and weird stuff happens. This branch adds "offline access" to the token scope, so the refresh token lasts a long time. It would let the user use the app all day without having to log back in after 5 or 6 hours. Currently still reading up on the consequences of "offline access"


duncan-oxd:
I don't think we wanna do this anymore. I tweaked some settings on the dev Keycloak to be closer to the defaults and/or production environment, and the offline stuff isn't necessary for a stable session anymore. Closing this in favor of #162 !

PR #138, opened 2025-02-25
CMS-650: Add error flash messages

duncan-oxd:
### Jira Ticket

CMS-650

### Description

This adds flash messages for incidental errors caused by calls to `fetchData` or `sendData`. In most of these places, we were just using console.error to print the error and there was no indication to the user. I left the console lines in, but added a red flash message.

I also left room for a third kind of flash message to alert the user when validation fails. (To be discussed with UX)

### DIFF
--- a/frontend/src/components/FlashMessage.jsx
+++ b/frontend/src/components/FlashMessage.jsx
@@ -23,6 +23,7 @@ function FlashMessage({
     ["error", faOctagonExclamation],
   ]);
 
+  // Allow custom icons, or use the variant / default
   const displayIcon = icon ?? variantIcons.get(variant) ?? faCheck;
 
   return (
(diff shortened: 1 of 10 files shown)


diego-oxd:
This would work and I'm not against merging it as is. 
But it'd be nice if we could everything in a single place. Especially, for all the errors related to API requests, we can add an interceptor to the axios instance, and have context that manages the state of when to show an error. Then, we wouldn't neex to import all these same stuff in every place. 
It would take some restructuring so I'm OK with merging this now and then having another ticket to clean all the flash messages. 
Up to you.

diego-oxd (approved):

duncan-oxd:
👍 Good plan, I want to have a "Flash message provider" at the root that provides the component and openFlashMessage() to all of the child components, so we don't need to add an individual FlashMessage to each component that uses it.

PR #144, opened 2025-02-27
NOBUG: Fix keycloak login issue

ayumi-oxd:
### Jira Ticket
None

### Description
- Fix an issue that the users couldn't log in to the Staff Portal by updating `keycloak-js`
- TODO: SP has a lot of outdated libraries and high vulnerabilities ... need to update them

### DIFF
--- a/frontend-legacy/package-lock.json
+++ b/frontend-legacy/package-lock.json
@@ -24,7 +24,7 @@
         "axios-mock-adapter": "^1.22.0",
         "history": "^5.0.0",
         "jsonwebtoken": "^8.5.1",
-        "keycloak-js": "^12.0.4",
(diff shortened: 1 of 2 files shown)


duncan-oxd (approved):
Going from version 12 to 26 could be risky, especially if @react-keycloak/web relies on it, but if everything else works after updating, then let's merge it!

ayumi-oxd:
> Going from version 12 to 26 could be risky, especially if @react-keycloak/web relies on it, but if everything else works after updating, then let's merge it!

Indeed it is risky, and unfortunately `@react-keycloak/web` is not supported anymore: https://www.npmjs.com/package/@react-keycloak/web ... but fortunately it works 😅

PR #146, opened 2025-02-27
CMS-735: Add missing dates and seasons

diego-oxd:
### Jira Ticket

CMS-735

### Description
- Added a script to create missing dates and seasons for 2025, this ensure that users can add dates for 2025 for every feature. Basically, the script will go through every feature:
    - check that a corresponding 2025 season exists for that feature, if it doesn't exist, create it
    - create an operation date for that feature-season if there's none
    - create  a reservation date for that feature-season if there's none and that feature allows for reservations
    - same process for winter fee seasons
 - when there's a previous dateRange that has null values for `startDate` or `endDate`, "Not submitted" will be displayed consistent with the wireframes. 
 - fixed small bug in which "not available" was displayed side by side with no spacing instead of different lines.
 
See the ticket for screenshots of where this is an issue.

### DIFF
--- a/backend/import-all-data.sh
+++ b/backend/import-all-data.sh
@@ -11,5 +11,6 @@ npm run sync-data
 npm run one-time-data-import
 npm run create-single-item-campgrounds
 npm run create-multiple-item-campgrounds
+npm run create-missing-dates-and-seasons
 
 echo "All commands executed successfully."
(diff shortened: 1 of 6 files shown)


duncan-oxd:
I hope multiple lines of "Not submitted" won't be an issue after we add the button to remove date ranges... but we'll still need to add something to clean up extra blank ranges when we do that 🤔

duncan-oxd (approved):
Great work!

PR #147, opened 2025-02-28
CMS-709: Add username and logout button to header

duncan-oxd:
### Jira Ticket

CMS-709
and part of CMS-634

### Description

There are a few overlapping tasks for the layout changes. Once we resolve this one, I'll review the remaining tasks and update them to have more focused requirements.

- Added a logout button to the header and removed the hidden logout route
(Moved the logout function into the access context)
- Added a `/user` endpoint to respond with the user object from the DB, and used that to show the user's name in the header
(This way the header will match the name in the changelogs, in case the user's name in DOOT is different from the name in the Keycloak token)
- Adjusted the header to be smaller, and changed the logo that displays on mobile to be the "vertical" BC Parks logo
- Tweaked the "link-text" button style to allow more than just the primary color.


diego-oxd:
Looks good, the only thing is that I think with the "staff web portal" text shouldn't be underlined. After that, should be good to go.

PR #150, opened 2025-03-05
CMS-694: Publish winter fee dates using new structure

diego-oxd:
### Jira Ticket

CMS-694

### Description
We are pushing Winter fee dates to the new collection in Strapi (park-feature-date) and operation/reservation dates to the old collection (park-operation-sub-area-dates).

When we are publishing a season, we might be publishing dates for many features. For all the features involved:
- every existing strapi date for that feature for that year will be set as `isActive: false` 
- the new dates will be created in their respective collection 
   - park-operation-sub-area-date for operation/reservation dates
   - park-feature-date for winter fee dates.


backend/routes/api/publish.js
```diff
@@ -202,13 +202,35 @@ async function createRecordsInStrapi(datesToPublish) {
...
+  const endpoint = "/api/park-feature-dates";
+
+  return Promise.all(
+    dates.map(async (date) => {
+      try {
+        const data = {
+          data: date,
+        };
+
+        await post(endpoint, data);
```
duncan-oxd, line 216: I think for this `map` to work, we should `return` the promise instead of `await`ing it here, otherwise it would just be an array full of `undefined` passed to the Promise.all

backend/routes/api/publish.js
```diff
@@ -280,84 +336,98 @@ async function markSeasonPublished(seasonId) {
...
-
-        await markFeatureDatesInactive(dates.data);
-
-        // The date object in strapi contains both the operating and reservation dates for a feature - operatingYear pair
-        // we'll group all the date ranges by feature and operating year and then we'll group them if possible
-        // If there are remaining dates, we'll create a new date object with the remaining dates
-        const operatingDates = dateRanges
-          .filter((dateRange) => dateRange.dateType.name === "Operation")
-          .sort((a, b) => new Date(a.startDate) - new Date(b.startDate));
-
```
duncan-oxd, line 302: I assume vscode took all of these blank lines out when changing the indent level or something; that happened to me the other day too. I'm definitely keen to keep them in if we can

duncan-oxd (changes requested):
Big script! A few very small notes, but overall it looks great

backend/routes/api/publish.js
diego-oxd: If we don't care about the result of the post request, do we still need to return anything? The async function returns a promise that would eventually resolve to undefined, but all we care about is that the Promise.all resolves once every function is done. 
I'm still reading up about it, but I don't think we need to return anything unless we care about the value it resolves to.

duncan-oxd: Oh! I guess not - I forgot that it implicitly returns a promise no matter what. It would still make sense to return the result from `post` here, so it would be more clear what what's making the promise resolve, but I guess it wouldn't make a difference if it's not being used.

diego-oxd: I'll add the return just for consistency with the other methods.

duncan-oxd (approved):

PR #151, opened 2025-03-05
CMS-657: run sync script from dashboard

diego-oxd:
### Jira Ticket

CMS-657

### Description
- added ability to run the sync script from the admin dashboard. However, instead of recreating the DB, it will just delete all the contents and re-run the sync script to import the data
- past seasons will default to "on API" instead of requested
- sync script will use the STRAPI_URL environment variable. This is needed so that we can test the publish workflow.
   - every time we import something and publish it to Strapi we'll publish to the equivalent environment. 
   - For this to work, we need the featureId (PKs and FKs) to match what's on the environment. 
   - this means the STRAPI_URL environment variable will be needed in each env and it will need to match the CMS API URL for that environment. `https://alpha-dev-cms.bcparks.ca, https://alpha-test-cms.bcparks.ca, https://dev-cms.bcparks.ca, https://test-cms.bcparks.ca`.
- removed individual files and commands from `package.json` and put it all into a single JS function to avoid duplicate code.

### DIFF
--- a/backend/strapi-sync/create-multiple-item-campgrounds.js
+++ b/backend/strapi-sync/create-multiple-item-campgrounds.js
@@ -517,8 +517,8 @@ async function createCampground(item) {
   );
 }
 
-export async function createCampgrounds(items) {
-  await Promise.all(items.map(async (item) => createCampground(item)));
+export async function createMultipleItemsCampgrounds() {
+  await Promise.all(campgrounds.map(async (item) => createCampground(item)));
 }
 
 // createCampgrounds(campgrounds);
--- a/backend/strapi-sync/create-multiple-item-campgrounds.js
+++ b/backend/strapi-sync/create-multiple-item-campgrounds.js
@@ -480,16 +480,16 @@ const campgrounds = [
 ];
 
 async function updateFeature(item, campgroundId) {
-  const strapiId = parseInt(item.featureId.split("_")[1], 10);
+  // const strapiId = parseInt(item.featureId.split("_")[1], 10);
 
   // get feature by featureId
-  // const feature = await getItemByAttributes(Feature, {
-  //   strapiFeatureId: item.featureId,
-  // });
-
   const feature = await getItemByAttributes(Feature, {
-    strapiId,
+    strapiFeatureId: item.featureId,
   });
+
+  // const feature = await getItemByAttributes(Feature, {
+  //   strapiId,
+  // });
 
   // set campgroundId and name
   feature.campgroundId = campgroundId;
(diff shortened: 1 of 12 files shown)


backend/strapi-sync/create-multiple-item-campgrounds.js
```diff
@@ -517,8 +517,8 @@ async function createCampground(item) {
   );
 }
 
-async function createCampgrounds(items) {
-  await Promise.all(items.map(async (item) => createCampground(item)));
+export async function createMultipleItemsCampgrounds() {
+  await Promise.all(campgrounds.map(async (item) => createCampground(item)));
```
duncan-oxd, line 521: Since `createCampground` just takes one parameter and it's the same one being passed by `.map`, we can simplify it by just passing `createCampground` itself as the callback

```suggestion
  await Promise.all(campgrounds.map(createCampground));
```

backend/strapi-sync/reset-and-import-data.js
```diff
@@ -0,0 +1,46 @@
...
+  await SeasonChangeLog.destroy({ where: {} });
+  await DateRange.destroy({ where: {} });
+  await Season.destroy({ where: {} });
+  await Feature.destroy({ where: {} });
+  await Campground.destroy({ where: {} });
+  await FeatureType.destroy({ where: {} });
+  await DateType.destroy({ where: {} });
+  await Park.destroy({ where: {} });
+  await Dateable.destroy({ where: {} });
+  await User.destroy({ where: {} });
```
duncan-oxd, line 30: It's probably about the same, but maybe the query is more optimized if we use destroyAll or truncate?
https://sequelize.org/docs/v7/querying/delete/#deleting-in-bulk

duncan-oxd (changes requested):
Looks good! It's a good plan to keep the migration stuff separate and just expose the "reset and re-import" stuff through AdminJS. 
I left a few comments about tidying up comments, but the rest looks great

backend/strapi-sync/reset-and-import-data.js
diego-oxd: I'll go with destroyAll. The docs say truncate is faster, but when trying truncate it will throw an error. You can only use it when there are not FK relations.

duncan-oxd: Fair enough! I bet truncate would only work if we carefully re-arranged the order and that would be tricky to maintain.

duncan-oxd (approved):

duncan-oxd:
I still need to add the STRAPI_* env vars to the other environments; I'll do that this morning ✅

PR #152, opened 2025-03-06
CMS-631: Update Edit/Preview pages

duncan-oxd:
### Jira Ticket

CMS-631

### Description

Big ticket with lots of changes on the Submit and Preview pages for winter and regular seasons. I should have split it into several branches 🙈 And I could still do that if it's too ridiculous to review. Here's a summary:

# Small stuff
There are screenshots in the ticket to clarify this stuff:
- Update the `NavBack` component text and link target to match what's currently in Figma
- Update validation styles: always show the icon, and always show red text on a white bg.
- Update spacing and header level for h1, h2, h3, h4's
- Update the header for regular seasons to match the new style already used on the Winter pages
- Remove fancy spacing for previous dates and just show plain text ("Previous: Fri, Mar 15...")
- Improve positioning of the dash between the date range inputs

# Ready to publish toggle
Use a CSS pseudo-element to add "On" and "Off" labels to the bootstrap toggle switch. Quick and dirty change, but seems fine.

# Add "X" button to remove extra rows of dates
- Hiding the button on the first date range in the list to prevent people from deleting all the dates for a feature, unless that's a requirement...
- Using an array of "deleted date range IDs" to track the ones to delete from the DB. Only changed ranges are sent in the payload, so it would be tricky to tell which ones are to delete or ignore otherwise. 
- No changes are made until you hit "Save" so you can just refresh or leave the page if you make a mistake, same as the other edit actions.
- I had to add an ON DELETE CASCADE constraint in a migration to delete the changelogs, but the backend changes are minimal.
- To get it to render right, I had to add a temporary ID for unsaved records because relying on key="index" won't work when you remove from the middle of the list.

# Full-width alternating row colors on Submit pages
The whole page was wrapped in a `container` class before. I had to move it into the content to wrap individual sections, but it wasn't too bad. Hopefully you can still make sense of the diff if you ignore whitespace.

### DIFF
--- a/backend/migrations/20250304210219-add-cascade-delete-to-datechangelogs.js
+++ b/backend/migrations/20250304210219-add-cascade-delete-to-datechangelogs.js
@@ -1,0 +1,41 @@
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    // Remove the existing constraint without CASCADE
+    await queryInterface.removeConstraint(
+      "DateChangeLogs",
+      "DateChangeLogs_dateRangeId_fkey",
+    );
+
+    // Add a new constraint with CASCADE
+    await queryInterface.addConstraint("DateChangeLogs", {
+      fields: ["dateRangeId"],
+      type: "foreign key",
+      name: "DateChangeLogs_dateRangeId_fkey",
+      references: {
+        table: "DateRanges",
+        field: "id",
+      },
+      onDelete: "CASCADE",
+    });
+  },
+
+  async down(queryInterface, Sequelize) {
+    // Remove the created constraint with CASCADE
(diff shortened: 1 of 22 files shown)


diego-oxd:
Great work. I just went through all the pages to make sure that what's mentioned in the ticket is covered and tried to look at the code as much as possible, but these differences in Github make it seem like more than it is. It all looks good. Only 2 things I found. 
- winter submit page has different spacing than regular season (dash between startDate and endDate)
- winter preview page uses h5 for campground names, it should be h4?? Not sure given the structure is different but it's h4 everywhere else.

duncan-oxd:
Good catch! Some of the changes didn't get copied over to the winter page. I pushed some small updates to fix those inconsistencies. I don't think there are supposed to be any H5s in the app at all anymore

PR #153, opened 2025-03-07
CMS-699: Optimize parks endpoints by reducing payload

diego-oxd:
### Jira Ticket

CMS-699

### Description
We don't need to fetch dates and features. This endpoint was created when other filters were needed and we didn't have a good idea of the payload size.


duncan-oxd (approved):
Nice! I'm glad the optimization was as easy as we'd hoped

PR #156, opened 2025-03-11
CMS-746: Allow users to continue to preview when dates are filled

diego-oxd:
### Jira Ticket

CMS-746

### DIFF
--- a/frontend/src/router/pages/SubmitDates.jsx
+++ b/frontend/src/router/pages/SubmitDates.jsx
@@ -101,6 +101,16 @@ function SubmitDates() {
     return notes;
   }
 
+  function continueToPreviewEnabled() {
+    const allDatesHaveValues = Object.values(dates).some((dateType) =>
+      dateType.Operation.concat(dateType.Reservation).every(
+        (dateRange) => dateRange.startDate && dateRange.endDate,
+      ),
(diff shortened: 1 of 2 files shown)


frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -101,6 +101,16 @@ function SubmitDates() {
     return notes;
   }
 
+  function continueToPreviewEnabled() {
```
duncan-oxd, line 104: Should we make this a reactive var with useMemo instead of a function? There's a ticket in the backlog to update hasChanges too.

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -101,6 +101,16 @@ function SubmitDates() {
...
+    const allDatesHaveValues = Object.values(dates).some((dateType) =>
```
duncan-oxd, line 105: Even though this array will only have one item (I think?), we should check with `.every` instead of `.some`
```suggestion
    const allDatesHaveValues = Object.values(dates).every((dateType) =>
```

duncan-oxd (changes requested):
Nice. A couple of tiny changes

PR #162, opened 2025-03-12
CMS-711: Attempt sign-in on auth errors

duncan-oxd:
### Jira Ticket

CMS-711

### Description

By signing in and out of all of the Keycloak consoles, I've experienced all kinds of weird and rare error states that shouldn't affect normal users. This branch adds a sign-in redirect to the auth error state, just in case something weird happens to anyone else.

Best case scenario: It will sign in and fix the token/cookies automatically

Worst case scenario: You'd go to the log-in page and have to enter your credentials again. Either way you wouldn't be stuck on an error page in DOOT.

Really worst case scenario would be if it got stuck in some kind of redirect loop, but I think you're more likely to see the keycloak login screen and stay there. I added some `console.error` lines to help test this out for a while.

### DIFF
--- a/frontend/src/router/ProtectedRoute.jsx
+++ b/frontend/src/router/ProtectedRoute.jsx
@@ -36,6 +36,8 @@ export default function ProtectedRoute({ children }) {
   }, [auth, hasTriedSignin]);
 
   if (auth.error) {
+    // If there's an error, redirect to the sign-in page
+    auth.signinRedirect();
     return <div>Authentication error: {auth.error?.message}</div>;
   }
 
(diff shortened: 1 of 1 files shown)


duncan-oxd:
I left my network tab/console open overnight and it looks like this doesn't work either. Maybe I'll make it do a sign _out_ attempt instead of sign _in_ and just leave it at that for this release 😷 😬

ayumi-oxd (approved):
LGTM! Haven't tried the token one but logging out manually worked. Just had a thought that it might be more user-friendly if users can go to the sign-in page from the log-out page at `/logged-out` (having a link or a header on that page). We can ask Amanda later if needed as a future improvement.

PR #163, opened 2025-03-13
CMS-757: Allow users to save incomplete dates as draft

diego-oxd:
### Jira Ticket

CMS-757

### Description
Validation will not be run when user clicks "save as draft". So users will be able to save dates with only field populated or clear both fields and those changes will be reflected in the DB. 

This applies only when saving as draft. When clicking "continue to preview" the validation wouldn't prevent that.

Apparently we need to start validating as soon as dates are entered. But that is captured in another ticket, so this ticket doesn't deal with it. 
https://bcparksdigital.atlassian.net/browse/CMS-756

Since requirements have been changing and they were never explicitly captured for saving data -- refer to https://apps.nrs.gov.bc.ca/int/confluence/display/BCPRS/Saving+and+submitting+dates

### DIFF
--- a/frontend/src/hooks/useValidation.js
+++ b/frontend/src/hooks/useValidation.js
@@ -235,20 +235,22 @@ export default function useValidation(dates, notes, season) {
     const operationExtent = dateExtents[dateableId].Operation;
     const reservationExtent = dateExtents[dateableId].Reservation;
 
-    // End date is one or more days after reservation end date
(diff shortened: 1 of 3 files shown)


duncan-oxd (approved):
You're right that this will be at odds with CMS-756 but that will be a larger change ina  future version, so these changes are good for now 👍

PR #166, opened 2025-03-18
CMS-756: Update validation: always check the whole form

duncan-oxd:
### Jira Ticket

CMS-756

### Description

Previously, we waited until the form was submitted before running validation, and we only showed one error at a time.

Now, we check all validation at all times, and show all the errors at once. Except the "Enter a start/end date" messages, which are bypassed until you try to submit the form.

To accomplish this:
- Run `validateForm` on every input in a `useEffect` (and use `useCallback` with all of the related functions so it works). I moved the other functions outside of the `useValidation` hook to keep them separate.
- Got rid of the specific `onUpdateDateRange` callback because `validateForm` runs on every change now.
- Instead of returning on the first error, check everything and keep track of the overall valid state with a boolean. Pass it up to `validateForm` so it can return true or false to prevent form submission like before.

Basically anything that was returning early is now checking all the fields, and then deciding if the form is valid or not by looking at the combined results.

We also changed the validation error icons in the fields and the red messages that appear. Restored some styles we previously deleted to show a message right above the submit buttons when submission is prevented by validation.

### DIFF
--- a/frontend/src/router/pages/SubmitDates.jsx
+++ b/frontend/src/router/pages/SubmitDates.jsx
@@ -2,9 +2,9 @@ import { useParams, useNavigate, Link } from "react-router-dom";
 import { useEffect, useState, useMemo } from "react";
 import { cloneDeep, set as lodashSet, omit } from "lodash";
 import {
-  faCircleInfo,
-  faTriangleExclamation,
   faCalendarCheck,
+  faCircleInfo,
+  faHexagonExclamation,
   faPlus,
   faXmark,
 } from "@fa-kit/icons/classic/regular";
@@ -42,7 +42,7 @@ function SubmitDates() {
   const { parkId, seasonId } = useParams();
 
   const [season, setSeason] = useState(null);
-  const [dates, setDates] = useState({});
+  const [dates, setDates] = useState({ pending: true });
   const [notes, setNotes] = useState("");
   const [readyToPublish, setReadyToPublish] = useState(false);
   // Track deleted date ranges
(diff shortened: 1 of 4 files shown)


frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -42,7 +42,7 @@ function SubmitDates() {
   const { parkId, seasonId } = useParams();
 
   const [season, setSeason] = useState(null);
-  const [dates, setDates] = useState({});
+  const [dates, setDates] = useState({ pending: true });
```
duncan-oxd, line 45: This is kinda sketchy, I dunno if we'd prefer to use a separate variable to track when the dates are loaded or something.

duncan-oxd: Actually yeah, I just updated it to default to null. Much more idiomatic and less weird!

frontend/src/router/pages/SubmitDates.jsx
```diff
@@ -101,21 +103,23 @@ function SubmitDates() {
...
+      ) && season?.status === "requested"
+    );
+  }, [dates, season]);
 
   async function saveChanges(savingDraft) {
-    setFormSubmitted(true);
+    formSubmitted.current = true;
 
     // Validate form state before saving
-    if (!savingDraft && !validateForm()) {
```
diego-oxd, line 118: You should be able to save drafts with incomplete dates according to https://bcparksdigital.atlassian.net/browse/CMS-757

duncan-oxd:
> if you open a season that is "approved" or "on API" you see the validation error for notes right away.

I'll double check the requirements around that, and push an update if needed. I'll tweak the check for savingDraft and make sure the incomplete dates validation only runs when you're continuing to the preview page.

diego-oxd (approved):

PR #174, opened 2025-03-25
CMS-588: backend role checks

diego-oxd:
### Jira Ticket

CMS-588

### Description
Added backend permission checks based on https://apps.nrs.gov.bc.ca/int/confluence/display/BCPRS/Roles?

The only checks missing are the ones based on bundles mapping users to specific parks. But there is another ticket in the backlog for that.

### DIFF
--- a/backend/middleware/permissions.js
+++ b/backend/middleware/permissions.js
@@ -10,10 +10,7 @@ export const adminsAndApprovers = ["doot-super-admin", "doot-approver"];
 export function checkPermissions(requiredRoles) {
   return (req, res, next) => {
     try {
-      if (
-        !req.auth?.resource_access ||
-        !req.auth.resource_access["staff-portal"]
-      ) {
+      if (!req.auth?.resource_access?.["staff-portal"]) {
         throw new Error("Authentication data missing");
       }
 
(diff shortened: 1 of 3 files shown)


backend/middleware/permissions.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {Object} req The request object.
+ * @param {Object} res The response object.
+ * @param {Function} next The next middleware function.
+ * @returns {Function} The next middleware function.
+ */
+export function sanitizePayload(req, res, next) {
+  try {
+    if (
+      !req.auth?.resource_access ||
+      !req.auth.resource_access["staff-portal"]
```
duncan-oxd, line 52: Can we combine these into a single check? Something like:
```suggestion
      !req.auth?.resource_access?.["staff-portal"]
```

backend/middleware/permissions.js
```diff
@@ -0,0 +1,80 @@
...
+    if (!allowedRoles.some((role) => userRoles.includes(role))) {
+      // check if readyToPublish is in the payload
+      if (req.body && "readyToPublish" in req.body) {
+        delete req.body.readyToPublish;
+      }
+
+      // remove status from the payload
+      if (req.body && "status" in req.body) {
+        delete req.body.status;
+      }
```
duncan-oxd, line 70: If we check req.body once, then I think we can just run those two `delete` lines without needing to check if they're set or not.

duncan-oxd (changes requested):
Very cool! I put in some thoughts about potentially simplifying the logic a bit, but overall it looks great

duncan-oxd (approved):

PR #181, opened 2025-03-27
CMS-772: Change text colour of disabled buttons

diego-oxd:
### Jira Ticket

CMS-772

### Description
Just the colour of disabled text buttons


duncan-oxd (approved):
I noticed the bootstrap theme dials down the opacity of disabled buttons to 65%, so the specific grey color here might be a little lighter than they want. Let's merge it, but we might also need to set the opacity to `1` later on.

PR #188, opened 2025-04-15
CMS-625: Add scroll anchor to winter edit page

duncan-oxd:
### Jira Ticket

CMS-625

### Description

I forgot to implement the scrolling logic on the Winter fee edit page in #185 so this is the exact same changes, but copied onto the Winter fee pages.


ayumi-oxd (approved):
Looks good!

PR #189, opened 2025-04-15
CMS-637: Add submitter role, "Pending review" state

duncan-oxd:
### Jira Ticket

CMS-637

### Description

This branch adds the "DOOT Submitter" user role. If a user has a "doot-submitter" role (and not a "doot-approver" or "doot-super-admin") they will see a few different elements in the UI:
- Removed buttons and wording for "Review" 
- Removed the "Approve" button and added a "Submit for approval" button
- Added a backend endpoint for the button above, which changes the season's status to "pending review" - mostly copied from the "approve" endpoint.
- On the frontend, we had a method to "save and approve" and we needed to add "save and submit" so I refactored them both to use "save and POST" and share most of their code.

Instead of specifically checking if the user is a submitter, the logic actually checks if the user _is not an approver_. There will be roles for RSOs and POs in the future, and it makes more sense to check for the approver role and show extra stuff, rather than check for the submitter role and selectively hide things. It's easier to make sense of the logic if we don't assume everyone is an approver, I think.

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -252,7 +252,7 @@ router.post(
   sanitizePayload,
   asyncHandler(async (req, res) => {
     const seasonId = Number(req.params.seasonId);
-    const { notes, dates, readyToPublish, deletedDateRangeIds = [] } = req.body;
+    const { notes, dates, deletedDateRangeIds = [] } = req.body;
 
     // when we add roles: we need to check that this user has permission to edit this season
     // staff or operator that has access to this park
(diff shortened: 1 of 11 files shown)


ayumi-oxd (commented):
LGTM! As a submitter, I confirmed some functionalities/buttons are not available, but I still see the Publish tab. It has been a while for me to run the DOOT app locally so I might have done wrong or haven't configured the role properly 😅 In that case please let me know!

> Hide publish tab for Submitters

duncan-oxd:
> Hide publish tab for Submitters

Good catch, I missed that in the requirements. Just pushed an update to hide it for non-approver users

ayumi-oxd (approved):
I don't see the publish tab as a submitter. LGTM :+1:

PR #190, opened 2025-04-16
CMS-759: Memoize hasChanges

duncan-oxd:
### Jira Ticket

CMS-759

### Description

This is a small change to turn the `hasChanges` function into a `useMemo` hook. I didn't think of it when I initially wrote that code. 😅  Hopefully it will cut down on unnecessary re-renders, or at least make the code a bit more maintainable. 

It's the same changes in the regular seasons and the winter seasons.

### DIFF
--- a/frontend/src/router/pages/WinterFeesSeasonPage.jsx
+++ b/frontend/src/router/pages/WinterFeesSeasonPage.jsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { Outlet, useNavigate, useParams } from "react-router-dom";
 import { omit } from "lodash-es";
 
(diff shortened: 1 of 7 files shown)


frontend/src/router/pages/WinterFeesSeasonPage.jsx
```diff
@@ -153,7 +153,9 @@ export default function WinterFeesSeasonPage() {
   }
 
   // Returns true if there are any form changes to save
-  function hasChanges() {
+  const hasChanges = useMemo(() => {
+    if (!dates) return false;
```
duncan-oxd, line 157: When it was an ordinary function, it wasn't called until after `dates` and `season` are set, but now it will be evaluated immediately when they're still null, so we just need to add this line to handle that case. The non-winter version already has it.

ayumi-oxd (approved):
I always forget to use useMemo ... looks great!

PR #191, opened 2025-04-16
CMS-648: Update react router and add v7 flags

duncan-oxd:
### Jira Ticket

CMS-648

### Description

A small change to just suppress some console warnings about features that will change in React router v7, even though we're using v6... They were annoying and they didn't affect our app, but they started showing when we upgraded from a previous version.

This branch updates to the latest 6.x.x version, and adds feature flags to opt in to the new features and hide the console warnings. Everything should still work the same because we're not using the new features.

### DIFF
--- a/frontend/src/main.jsx
+++ b/frontend/src/main.jsx
@@ -13,7 +13,13 @@ createRoot(document.getElementById("root")).render(
   <StrictMode>
     <AuthProvider onSigninCallback={onSigninCallback} {...oidcConfig}>
       <ProtectedRoute>
-        <RouterProvider router={router} />
+        <RouterProvider
+          router={router}
+          future={{
+            // eslint-disable-next-line camelcase -- vendor flag
+            v7_startTransition: true,
+          }}
+        />
       </ProtectedRoute>
     </AuthProvider>
   </StrictMode>,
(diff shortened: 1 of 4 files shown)


frontend/src/main.jsx
```diff
@@ -13,7 +13,13 @@ createRoot(document.getElementById("root")).render(
   <StrictMode>
     <AuthProvider onSigninCallback={onSigninCallback} {...oidcConfig}>
       <ProtectedRoute>
-        <RouterProvider router={router} />
+        <RouterProvider
+          router={router}
+          future={{
+            // eslint-disable-next-line camelcase -- vendor flag
+            v7_startTransition: true,
+          }}
```
ayumi-oxd, line 21: What does it do? Just a flag?

ayumi-oxd (approved):

ayumi-oxd:
Just a reminder, we also have `react-router-dom` `v5.2.0` in the legacy app :open_mouth:

duncan-oxd:
I didn't look into it, but it would probably be not too much work to upgrade the legacy app from 5.x to the current 7.x (and DOOT from 6.x to 7.x) Unless we're doing something special. I bet for a typical router config, it should be pretty straight forward to upgrade...

PR #193, opened 2025-04-18
NOBUG: Update pr template

ayumi-oxd:
### Jira Ticket
None

### Description
- I feel like we always delete one line above "CMS-" in the comment but please let me know if it's an important thing 😄

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,5 @@
 ### Jira Ticket
-
 CMS-
 
 ### Description
 <!-- What did you change, and why? -->
-


duncan-oxd:
It actually renders the same with or without a blank line after a heading, it's just a formatting thing (markdown linters will automatically add it) For example, if you open up one of the README files in a dev container and delete a blank line after a heading and hit save, it will re-add the blank line, and the preview pane won't change.

I'd prefer leaving them in! Now that I'm comparing to the template in the main repo, I notice that one has ":" after the headings. We could take those out of the other template for consistency if you want :D 

## Example with a line between

Hello

## Example with no line between
Hello

ayumi-oxd:
Good to know there's no difference in the end. I always deleted the line haha 🤦🏻‍♀️

PR #194, opened 2025-04-23
CMS-606: Transform date strings in useApi

duncan-oxd:
### Jira Ticket

CMS-606

### Description

We currently convert between date strings and date objects in several places when dates are stored or manipulated. This branch transforms all dates to date objects when the data is loaded (in the Axios instance) and then the dates are objects for the whole lifecycle of the app until they're sent back to the API. This will reduce complexity where we had to convert formats (and in some cases, time zones).

The changes involve mostly removing places where ISO date strings are parsed, and overall it simplifies things a bit. 

A future change we could make would be to store the date ranges as DATEONLY types in the db, so they wouldn't have the hour part and suffer from time-zone complications. But that can happen later on if we decide it will be helpful.

### DIFF
--- a/frontend/src/hooks/useValidation.js
+++ b/frontend/src/hooks/useValidation.js
@@ -1,19 +1,16 @@
 import { useEffect, useState, useCallback, useMemo, useRef } from "react";
-import { omit, mapValues, minBy, maxBy, orderBy } from "lodash-es";
-import { differenceInCalendarDays, parseISO, isBefore, max } from "date-fns";
+import { omit, mapValues, minBy, maxBy, orderBy, cloneDeep } from "lodash-es";
+import { differenceInCalendarDays, isBefore, max } from "date-fns";
 
 // Returns a chronological list of date ranges with overlapping ranges combined
 function consolidateRanges(ranges) {
-  // Parse ISO date strings (and filter out any missing values)
-  const parsedRanges = ranges
-    .filter((range) => range.startDate && range.endDate)
-    .map((range) => ({
-      startDate: parseISO(range.startDate),
-      endDate: parseISO(range.endDate),
-    }));
+  // Filter out any missing values
+  const filteredRanges = ranges.filter(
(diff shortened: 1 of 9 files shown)


frontend/src/hooks/useValidation.js
```diff
@@ -1,19 +1,16 @@
...
-      endDate: parseISO(range.endDate),
-    }));
+  // Filter out any missing values
+  const filteredRanges = ranges.filter(
+    (range) => range.startDate && range.endDate,
+  );
 
   // Sort ranges by start date
-  const sorted = orderBy(parsedRanges, ["startDate"]);
+  const sorted = orderBy(cloneDeep(filteredRanges), ["startDate"]);
```
duncan-oxd, line 13: using cloneDeep here prevents the code below from modifying the original dates object. Even if we scrap this branch, I'd still want to keep that change, just to make the code more solid.

ayumi-oxd: Cool! I didn't know lodash has this function. What would be the reason you use this `cloneDeep` in here and not in other places? Will it modify the data if you use `orderBy`?

ayumi-oxd (approved):
LGTM!

frontend/src/hooks/useValidation.js
duncan-oxd: It's very handy if you have nested objects, because copying an object with `{ ...myObject }` or `Object.assign({}, myObject)` could still cause unintended side-effects if you mutate nested objects inside. 
Deep-cloning a whole object can be expensive to do, and usually indicates that there's something inefficient happening earlier on, so I treat it as a last resort 😝 

`orderBy` returns a new object, so the original `filteredRanges` variable isn't modified, but since it has other objects nested inside it, they still mutate the original object when we change `lastRange.endDate` etc. 

Once we arrive at our final data structure, I'll put a ticket in the backlog to refactor and optimize the validation code, but it's going to all change in the near future anyway 🙃

PR #199, opened 2025-05-09
CMS-853: Rename Campground model to ParkArea

duncan-oxd:
### Jira Ticket

CMS-853

### Description

This is the first step of 2 (or more) to update the data model for DOOT v2. This one renames the "Campground" taxonomy to "ParkArea" to avoid confusion, and future-proof us for when we have more than just campsite features grouped into these areas.

I also updated the backend Readme file, which was slightly out of date with the npm scripts listed.


backend/models/featuretype.js
```diff
@@ -15,7 +15,7 @@ export default (sequelize) => {
 
       FeatureType.hasMany(models.Season, {
         foreignKey: "featureTypeId",
-        as: "campgrounds",
+        as: "seasons",
       });
```
ayumi-oxd, line 19: Why was it `campgrounds` before? I'm curious how it worked, or was it just a bug? 🙂

ayumi-oxd (approved):
Migration worked as expected! LGTM 👍

backend/models/featuretype.js
duncan-oxd: I meant to leave a comment on this PR about that; I'm pretty sure it was a bug before but we were never using it in a query so it didn't come up 🤔

PR #204, opened 2025-05-16
CMS-850: Add 'first come, first served' calculation

ayumi-oxd:
### Jira Ticket

CMS-850

### Description
- Add 'first come, first served' calculation based on the operation date and reservation date 
- FCFS dates will be calculated and approved when the season is approved
- FCFS dates will be sent to Strapi `park-feature-date` when the data is published (on API) 
- TODO: Test it again when approving dates are available again
- TODO: Publish not only winter dates but also other dates to Strapi `park-feature-date`
- TODO: Check if FCFS dates will be sent to Strapi too

### DIFF
--- a/backend/routes/utils/dateHelper.js
+++ b/backend/routes/utils/dateHelper.js
@@ -1,0 +1,83 @@
+import { DateRange, DateType } from "../../models/index.js";
+
+// FCFS dates calculation helper function
+
+// Example 1: Operation: 2025-03-21 to 2025-10-13
+//            Reservation: 2025-03-28 to 2025-10-12
+//            First come: 2025-03-21 to 2025-03-27
+// Example 2: Operation: 2025-04-01 to 2025-10-31
+//            Reservation: 2025-05-10 to 2025-10-12
+//            First come: 2025-04-01 to 2025-05-09
+//            First come: 2025-10-13 to 2025-10-31
+
+/**
+ * @param {Object} season
+ */
+
+export async function createFirstComeFirstServedDateRange(season) {
+  // Check if the FeatureType is "Frontcountry camping" or "Walk-in camping"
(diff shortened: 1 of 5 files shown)


ayumi-oxd:
It's ready for the review, but I'll merge it after the date types change. Also, the testing by QA isn't really doable since it needs the new publish flow... 😓

backend/routes/utils/dateHelper.js
duncan-oxd, line 1: Let's move this utils directory out to the root of /backend so everything in /routes will just be route files. (Perhaps some day we'll update it to automatically import all the routes in the directory, for example)

And let's call this file `firstComeFirstServed` or something, so it can just contain helpers for FCFS date stuff 👍  That will prevent us from making the file get big and disorganized with random functions, which always happens 😆

backend/routes/utils/dateHelper.js
```diff
@@ -0,0 +1,86 @@
...
+        firstComeStartDate.setDate(firstComeStartDate.getDate() + 1);
+
+        // Subtract one day from operation end date
+        const firstComeEndDate = new Date(operationDateRange.endDate);
+
+        firstComeEndDate.setDate(firstComeEndDate.getDate() - 1);
+
+        firstComeDateRanges.push({
+          seasonId: season.id,
+          dateTypeId: dateTypeMap["First come, first served"],
```
duncan-oxd, line 74: That reminds me we will probably have to add a separate "display name" field to the DateTypes table, so we won't have to go and update everything if they change these names. I'll put it in the backlog!

duncan-oxd: CMS-961 created for this 👍

duncan-oxd (approved):
Looks good!

PR #205, opened 2025-05-20
CMS-926: Add new date types

ayumi-oxd:
### Jira Ticket

CMS-926

### Description
- Add `level` to `dateType` to define which level it should be used at
- Change `description` string to text so that it can have more than 255 characters
- Add new date types to `dateType`
- Update `createOrUpdateDateType` function to update date types often rather than creating at once
- Note: This PR can go first before the FCFS PR


backend/migrations/20250519221250-add-level-field-to-datetype.js
```diff
@@ -0,0 +1,14 @@
...
+      type: Sequelize.ENUM("park", "feature"),
+      allowNull: false,
+      defaultValue: "feature",
+    });
+  },
+  async down(queryInterface, Sequelize) {
+    await queryInterface.removeColumn("DateTypes", "level");
+    await queryInterface.sequelize.query('DROP TYPE IF EXISTS "enum_DateTypes_level";');
+  },
+};
```
ayumi-oxd, line 14: A date type with the level "feature" can be used at the "park area" level too. I'm not sure if it'd be the best way so open to your thoughts 🙂

duncan-oxd: I wondered about this too. I almost think it might be easier to add separate field to indicate:
can it be used at the park level? yes/no
can it be used at the park area level? yes/no
can it be used at the feature level? yes/no

That way, we can also correctly classify date types that can be used at more than one level. (Which I think will definitely happen)

How about three boolean fields instead of an enum here? parkLevel, parkAreaLevel, featureLevel or something?

backend/strapi-sync/date-types.js
```diff
@@ -0,0 +1,79 @@
...
+  },
+  {
+    name: "Operation",
+    startDateLabel: "Service Start Date",
+    endDateLabel: "Service End Date",
+    description:
+      "For areas managed by a park operator, this includes all dates when the PO is operating. For non-reservable backcountry and marine-access campgrounds, this includes all dates when camping is allowed, whether a PO is operating or not. For areas not managed by POs, this includes all times when the area is available for public use, and access is not blocked by a gate.",
+    level: "feature",
+  },
+  {
```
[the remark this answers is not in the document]
duncan-oxd: It will affect some stuff, but we could fix it with some find-and-replace operations. I think we should do that in a separate branch though because it would change a lot of files. We're already displaying Operating/Operation differently on the frontend so we can probably leave it for now in the DB and just tweak how it's shown to the user.

It would make sense to change it eventually, though...

backend/strapi-sync/date-types.js
```diff
@@ -0,0 +1,79 @@
+export const dateTypesData = [
+  {
+    name: "Backcountry registration",
+    startDateLabel: "Backcountry registration start date",
+    endDateLabel: "Backcountry registration end date",
+    description:
+      "Enter dates when backcountry permit registration is required either online or at self-registration kiosks. If required year-round, enter Jan 1 to Dec 31.",
+    level: "feature",
```
duncan-oxd, line 8: With my other suggestion, we could make this an array. 
Right now, there is quite a bit of overlap between ParkArea and Feature-level dates and I'd like to keep the system flexible enough to handle any combination of requirements for where and how dates get used.

We can still have two date type records with the same name if we want to have different descriptions at different levels (like with Winter fees at the bottom of this file)

backend/strapi-sync/sync.js
```diff
@@ -258,33 +259,29 @@ export async function syncFeatureTypes(strapiData, featureTypeData) {
...
+export async function createOrUpdateDateType(item) {
+  let dbItem = await DateType.findOne({
+    where: { name: item.name, level: item.level },
+  });
 
-  await Promise.all(data.map((item) => createModel(DateType, item)));
+  if (dbItem) {
+    dbItem.startDateLabel = item.startDateLabel;
+    dbItem.endDateLabel = item.endDateLabel;
+    dbItem.description = item.description;
```
duncan-oxd, line 274: It doesn't make a huge difference for only 3 fields, but I just noticed there's a `set` function for updating multiple fields that could be useful here: https://sequelize.org/docs/v6/core-concepts/model-instances/#updating-an-instance

backend/strapi-sync/sync.js
```diff
@@ -258,33 +259,29 @@ export async function syncFeatureTypes(strapiData, featureTypeData) {
 
 /**
  * These are not associated with any specific model in Strapi
- * We will create these manually once
- * @returns {Promise[Object]} resolves when all dateTypes have been created
+ * We will create these manually
+ * @param {Object} item dateType data from db
+ * @returns {DateType} dateType model
```
duncan-oxd, line 264: Since it's an async function, it technically returns a Promise that resolves to a DateType

I think there's a mix of angle brackets and square brackets in this project, but it's mostly angle brackets so let's go with that here:

```suggestion
 * @returns {Promise<DateType>} dateType model
```

duncan-oxd (changes requested):
It's on the right track! I put in a suggestion about using 3 boolean fields instead of an enum -- check it out and let me know if it makes sense or if there's any reason why that wouldn't be a good idea. 

I keep running into situations where a JSONB array field would do the trick, but I'm trying to do everything the SQL-ish way 😆

PR #214, opened 2025-06-05
CMS-894: Update park API endpoint

ayumi-oxd:
### Jira Ticket

CMS-894

### Description
- Get all of `parkArea` in `park`
- Get `feature` in `parkArea` if it has `parkAreaId`
- Get `feature` in `park` if it doesn't have `parkAreaId`
- Get `season` in `park`/`parkArea`/`feature` if its `publishableId` is matched
- Add `readyToPublish` from `season` to custom grouped dates to display a flag
- Code clean up

### DIFF
--- a/backend/models/feature.js
+++ b/backend/models/feature.js
@@ -32,6 +32,12 @@ export default (sequelize) => {
         foreignKey: "publishableId",
         as: "publishable",
       });
+
+      Feature.hasMany(models.Season, {
+        foreignKey: "publishableId",
+        sourceKey: "publishableId",
+        as: "seasons",
+      });
     }
   }
   Feature.init(
--- a/backend/models/feature.js
+++ b/backend/models/feature.js
@@ -35,7 +35,9 @@ export default (sequelize) => {
 
       Feature.hasMany(models.Season, {
         foreignKey: "publishableId",
-        sourceKey: "publishableId",
+        // Features that are part of the ParkArea should be displayed even if the publishableId is null
+        // e.g. Shuswap Lake Campground has features site 1-78 and site 79-245
+        // sourceKey: "publishableId",
         as: "seasons",
       });
     }
--- a/backend/models/feature.js
+++ b/backend/models/feature.js
@@ -35,9 +35,7 @@ export default (sequelize) => {
 
       Feature.hasMany(models.Season, {
         foreignKey: "publishableId",
-        // Features that are part of the ParkArea should be displayed even if the publishableId is null
-        // e.g. Shuswap Lake Campground has features site 1-78 and site 79-245
-        // sourceKey: "publishableId",
+        sourceKey: "publishableId",
         as: "seasons",
       });
     }
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -14,11 +14,9 @@ import asyncHandler from "express-async-handler";
 
 // Constants
 const router = Router();
-const currentYear = new Date().getFullYear();
-const minYear = currentYear - 1;
 
 // Functions
-function seasonModel(required = true) {
+function seasonModel(minYear, required = true) {
   return {
     model: Season,
     as: "seasons",
@@ -54,7 +52,7 @@ function seasonModel(required = true) {
   };
 }
 
-function featureModel(where = {}) {
+function featureModel(minYear, where = {}) {
   return {
     model: Feature,
     as: "features",
@@ -75,46 +73,21 @@ function featureModel(where = {}) {
         attributes: ["id", "publishableId", "name"],
       },
       // Publishable Seasons for the Feature
-      seasonModel(false),
+      seasonModel(minYear, false),
     ],
   };
 }
 
 // get publishable item status from seasons
 function getStatus(seasons) {
-  // @TODO: fix this
-  if (!seasons) return "requested";
+  const currentYear = new Date().getFullYear();
+  // filter seasons by current year
+  const currentSeasons = seasons.filter(
+    (season) => season.operatingYear === currentYear,
+  );
 
-  // if any season has status==requested, return requested
-  // else if any season has status==pending review, return pending review
-  // else if any season has status==approved, return approved
-  // if all seasons have status==on API, return on API
-
-  const requested = seasons.some((s) => s.status === "requested");
-
-  if (requested) {
-    return "requested";
-  }
-
-  const pendingReview = seasons.some((s) => s.status === "pending review");
-
-  if (pendingReview) {
-    return "pending review";
-  }
-
-  const approved = seasons.some((s) => s.status === "approved");
-
-  if (approved) {
-    return "approved";
-  }
-
-  const onAPI = seasons.some((s) => s.status === "on API");
-
-  if (onAPI) {
-    return "on API";
-  }
-
-  return null;
+  // return status of the first current season
+  return currentSeasons.length > 0 ? currentSeasons[0].status : null;
 }
 
 // group dateRanges by date type name then by year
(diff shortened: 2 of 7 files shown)


backend/models/feature.js
```diff
@@ -32,6 +32,14 @@ export default (sequelize) => {
         foreignKey: "publishableId",
         as: "publishable",
       });
+
+      Feature.hasMany(models.Season, {
+        foreignKey: "publishableId",
+        // Features that are part of the ParkArea should be displayed even if the publishableId is null
+        // e.g. Shuswap Lake Campground has features site 1-78 and site 79-245
+        // sourceKey: "publishableId",
```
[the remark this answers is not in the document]
ayumi-oxd: Adding `sourceKey: "publishableId"` filters `feature` that doesn't have `publishableId`, but we still need to display it in the table.

backend/routes/api/parks.js
```diff
@@ -5,18 +5,58 @@ import {
...
   ParkArea,
 } from "../../models/index.js";
 import asyncHandler from "express-async-handler";
 
+// Constants
 const router = Router();
+const currentYear = new Date().getFullYear();
+const minYear = currentYear - 1;
+
+// @TODO: it is not repeatable?
```
ayumi-oxd, line 20: I was able to use it in one place ... so I just added the same structures repeatedly at other levels 😓
It'd be great to know if there is another way not to repeat it.

backend/routes/api/parks.js
```diff
@@ -5,18 +5,58 @@ import {
...
+  ],
+};
 
 // Functions
-function getParkStatus(seasons) {
+// get publishable item status from seasons
+function getStatus(seasons) {
+  // @TODO: fix this
+  if (!seasons) return "requested";
+
```
ayumi-oxd, line 59: What is TODO here?

duncan-oxd: I think we don't need a function to figure out a calculated status anymore: every season just has its own status field that we can red directly

backend/routes/api/parks.js
duncan-oxd: I guess the Season, DateRange, DateType variables get mutated during a query so we can't reuse query chunks as variables, but we _can_ create a function that returns a query chunk 👍 I sent a message on slack about it

backend/routes/api/parks.js
```diff
@@ -164,19 +225,106 @@ router.get(
...
+              model: Feature,
+              as: "features",
+              attributes: [
+                "id",
+                "dateableId",
+                "publishableId",
+                "parkAreaId",
+                "name",
+                "inReservationSystem",
+              ],
```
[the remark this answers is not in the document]
ayumi-oxd: Added `required: false` in `featureModel` but didn't help the `sourceKey` thing 🥲

backend/models/feature.js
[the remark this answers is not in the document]
ayumi-oxd: It looks like Sequelize associates `Feature.publishableId` with `Season.publishableId` and returns if they are matched.
https://sequelize.org/docs/v6/core-concepts/assocs/#for-hasone-and-hasmany-relationships

backend/routes/api/parks.js
duncan-oxd: I don't know about the requirements for this part - if I put that `TODO` in when I was testing things, we can just undo it and grab the original code back from the alpha branch, or whatever we need to do. 
My understanding was that any level that shows a publishing status badge ("Requested", "On Api", etc.) will also have an "Edit" button right next to it, and those will both come directly from the season for the current year, so I thought we could just read the publishing status from the season record. But maybe that's not correct 🙃 
Feel free to ditch my changes if they're wrong!

backend/models/feature.js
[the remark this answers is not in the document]
duncan-oxd: That's what it should be doing - if a feature has no publishableId then it can't have any Seasons, so you'd have to query for features some other way.
(I think removing that line makes the association do nothing, so we probably just need to tweak a query somewhere, let me know if you want to chat about this!)

backend/routes/api/parks.js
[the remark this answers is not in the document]
duncan-oxd: Try adding it on the "season" include inside of the Feature block 👀

duncan-oxd:
I should also add that I get an error on the main page with this branch from this line in the table component:
```jsx
typeName={parkArea.featureType.name}
```
but maybe I need to do something with my data? Something else we can discuss tomorrow 😆

backend/routes/api/parks.js
ayumi-oxd: I'll ask Amanda so we can visit here again.

ayumi-oxd:
> I should also add that I get an error on the main page with this branch from this line in the table component:
> 
> ```js
> typeName={parkArea.featureType.name}
> ```
> 
> but maybe I need to do something with my data? Something else we can discuss tomorrow 😆
> 
> > Sorry, an unexpected error has occurred.
> > _Cannot read properties of null (reading 'name')_

Prob it didn't get `feature` in `parkArea` properly. Added `?` for quick fix. I couldn't recreate this issue on my end but let me know if you find something 🙏

backend/routes/api/parks.js
```diff
@@ -5,48 +5,89 @@ import {
...
-  if (onAPI) {
-    return "on API";
-  }
+// get publishable item status from seasons
+function getStatus(seasons) {
+  const currentYear = new Date().getFullYear();
+  // filter seasons by current year
+  const currentSeason = seasons.find(
+    (season) => season.operatingYear === currentYear,
+  );
```
duncan-oxd, line 87: We will need the current season's ID for many things in the frontend, so how about:
- Find the current Season in the main park endpoint, and just pass in that one season to get the status (or maybe we don't need the function at all at that point?)
- Return `currentSeasonId` or something with the Park details so it can be used for the edit/approve buttons in the table

duncan-oxd: all of the park areas in `park.parkAreas` and the features in `park.features` will need their `currentSeasonId` too: anywhere that shows a status/edit/approve buttons

ayumi-oxd: Added `currentSeason` object to each level so we can use its status and id.

backend/routes/api/parks.js
```diff
@@ -125,14 +150,48 @@ function buildFeatureOutput(feature, allDateRanges, parkSeasons) {
...
+  return {
+    id: parkArea.id,
+    dateableId: parkArea.dateableId,
+    publishableId: parkArea.publishableId,
+    name: parkArea.name,
+    features: parkArea.features.map((feature) => buildFeatureOutput(feature)),
+    featureType: featureType ?? null,
+    seasons: parkArea.seasons,
+    status: getStatus(parkArea.seasons),
+    groupedDateRanges: groupDateRangesByTypeAndYear(parkAreaDateRanges),
```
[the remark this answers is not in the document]
duncan-oxd: I added a date for the feature's current season and it does work 👍 
The date I had in there before was for the Park's season - and after our discussion yesterday, I don't think that scenario will ever happen in this version, so we don't need to handle it. Looks good! I'll resolve this thread ✅

backend/routes/api/parks.js
ayumi-oxd: It can be multiple dated in the dateRanges array, but we use `groupedDateRanges` for the table display anyway so we can remove it from `currentSeason`. For the FE, only `status` is required for now but we can customize it later if needed.

ayumi-oxd: Removed `currentSeason` from Feature in ParkArea. It'll be needed at the other levels, Park, Park.ParkArea, Park.Feature to display its status.

duncan-oxd (approved):
Great work on this big ticket! It looks like it's working the way I'd expect on my local machine, but I don't have a lot of dates in my DB 😅  If any bugs are found, we can address them in the future, but I think this is good to merge now 🎉

PR #216, opened 2025-06-10
CMS-930: Update Season endpoints for v2 data model

duncan-oxd:
### Jira Ticket

CMS-930

### Description

This updates the /season/ endpoints to work with the new data structure. For the GET endpoints, there are 3 separate ones (park-level, park-area-level, and feature-level) because the structure is different.

In every case, it returns the season object, with nested objects for:
- park / feature / parkArea objects inside it
- dateable IDs inside the objects
- dateRanges inside the dateables

park would include parkAreas, parkAreas would include features, etc.

The /save/ endpoint is pretty much the same as v1: we'll send a payload with a flat list of DateRanges:
- dateRange with an ID will be updated
- dateRange with no ID will be inserted
- IDs in the `deletedDateRangeIds` array will be deleted

Still to do:

- [ ] use transactions for endpoints with more than one db mutation


backend/routes/api/seasons.js
```diff
@@ -23,474 +24,490 @@ import {
...
+// Common attributes for all Season queries
+const SEASON_ATTRIBUTES = [
+  "id",
+  "operatingYear",
+  "status",
+  "readyToPublish",
+  "editable",
+  "publishableId",
+];
+
```
ayumi-oxd, line 190: Useful!

backend/routes/api/seasons.js
```diff
@@ -23,474 +24,490 @@ import {
...
+              as: "parkArea",
               attributes: ["id", "name"],
+              required: false,
+            },
+
+            // Park details
+            {
+              model: Park,
+              as: "park",
             },
```
ayumi-oxd, line 219: Not sure if we need ParkArea and Park inside Feature since we have a form at each level. But if not, we can edit the model anytime 🙂

backend/routes/api/seasons.js
```diff
@@ -23,474 +24,490 @@ import {
...
+                  as: "features",
+                  include: [
+                    featureTypeQueryPart(),
+
+                    // Dates for this Feature Season
+                    dateableAndDatesQueryPart(seasonId),
+                  ],
+                },
+              ],
+            },
```
ayumi-oxd, line 331: Same as above. Does Park need Park Area and Feature? but we can modify API anytime.

backend/routes/api/seasons.js
```diff
@@ -23,474 +24,490 @@ import {
...
 
-      error.status = 404;
-      throw error;
-    }
+    const transaction = await sequelize.transaction();
 
-    // Create "First come, first served" DateRange if applicable
-    await createFirstComeFirstServedDateRange(season);
+    try {
+      await updateStatus(seasonId, "approved", null, transaction);
```
ayumi-oxd, line 487: Not necessary but it'd be nice if we had status enum values somewhere. Then we can use it like `SEASON_STATUS.APPROVED`. It's because I make typos and forget if the first letter is uppercase/lowercase 😅

```js
const SEASON_STATUS = {
  REQUESTED: "requested",
  APPROVED: "approved",
  PENDING_REVIEW: "pending review",
  // more statuses
};
```

ayumi-oxd (approved):
Looks good! Just had some questions, but we can update this endpoint in CMS-896 or CMS-909 if needed.

Also I'll update the FCFS sometime soon!

backend/routes/api/seasons.js
```diff
@@ -23,474 +24,490 @@ import {
...
+      where: {
+        operatingYear: currentSeason.operatingYear - 1,
+        publishableId: currentSeason.publishableId,
+      },
+      include: [
+        {
+          model: DateRange,
+          as: "dateRanges",
+          required: false,
+        },
```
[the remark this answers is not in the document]
duncan-oxd: Pushed an update to include the dateType object with name and id 👍

PR #219, opened 2025-06-16
CMS-792: Check if a date is empty

ayumi-oxd:
### Jira Ticket

CMS-792

### Description

- Looks like it returned the Unix epoch time if a date parameter is empty, so change it to return an empty string


duncan-oxd (approved):
Looks good 👍 The January 1st date was probably confusing for the testers who saw it 😅 
There are future tickets to update this form to update the query for the v2 data structure, so this is ok to merge now ✅

PR #220, opened 2025-06-17
CMS-872: Add isDateRangeAnnual to all levels

ayumi-oxd:
### Jira Ticket

CMS-872


backend/migrations/20250616222440-add-is-date-range-annual-field-to-all-levels.js
duncan-oxd, line 1: The requirements say that there will be a checkbox for every date type:

> There is one check box per date type per Park, Area, and Feature

So Operating dates might be annual, but Reservation dates on the same Feature might not be 😬 I think that means it won't work to have one field for the whole model. 

Any ideas how this could work? Maybe we need an "AnnualDates" table that connects a dateTypeId to a publishableId?

backend/models/daterangeannual.js
```diff
@@ -0,0 +1,50 @@
...
+        references: {
+          model: "Publishables",
+          key: "id",
+        },
+      },
+      isDateRangeAnnual: {
+        type: DataTypes.BOOLEAN,
+        allowNull: false,
+        defaultValue: false,
+      },
```
duncan-oxd, line 42: I was thinking if the record is in this table at all, then we can consider the date range to be annual. Anything that would be "false" could simply be left out of this table. Would that work? If so, we can remove this column

backend/tasks/create-date-range-annual/create-date-range-annual.js
```diff
@@ -0,0 +1,168 @@
...
+  try {
+    // get dateType "Operating"
+    const operatingDateType = await DateType.findOne({
+      where: { name: "Operating" },
+      transaction,
+    });
+
+    if (!operatingDateType) {
+      throw new Error('No DateType with name "Operating" found.');
+    }
```
duncan-oxd, line 34: Do we need to specify if this is the park-level one or the other one?

duncan-oxd (commented):
Looks good! I had a thought about a way to (possibly) simplify it. If the current setup is easier to sync and keep it up to date, then we don't have to change it. We'd just be selecting "where isDateRangeAnnual = true" anyway, so the false records would never get used, except in the sync script to prevent duplicates. (I think?)

If we wanted to drop the `false` rows, we could just clear out the table every time the script runs and only insert the `true` ones.

backend/models/daterangeannual.js
ayumi-oxd: In that case, when a user submits data with `isDateRangeAnnual` checkbox from FALSE to TRUE, the Season API doesn't have DateRangeAnnual anywhere. So we'll need to insert a new row into the DateRangeAnnual when the Season is submitted? And after that, if a user changes it to FALSE, that row should be dropped? Double-checking if I understood correctly 😅

backend/tasks/create-date-range-annual/create-date-range-annual.js
ayumi-oxd: This is for the Park-level: https://bcgov.sharepoint.com/:x:/r/teams/031072/_layouts/15/Doc2.aspx?action=edit&sourcedoc=%7B9750a64c-a720-480a-8aee-0f0596150040%7D&wdOrigin=TEAMS-WEB.teamsSdk_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1742327891304&web=1 so used it in the parks iteration (line: 123).

```js
    // 2 - create dateRangeAnnual for "Operating" dateType, for each park
    for (const park of parks) {
      // get isDateRangeAnnual from parkOperationData using orcs
      const isDateRangeAnnual = parkOperationIsAnnualByOrcs[park.orcs] ?? false;

      const [entry, created] = await DateRangeAnnual.findOrCreate({
        where: {
          publishableId: park.publishableId,
          dateTypeId: operatingDateType.id,
        },
        defaults: {
          publishableId: park.publishableId,
          dateTypeId: operatingDateType.id,
          isDateRangeAnnual,
        },
        transaction,
      });
      ...
```

backend/models/daterangeannual.js
duncan-oxd: Ahh, you're right... that would be kind of messy to have to add and delete rows every time the user updates the value. Let's leave it the way it is, and most of the queries can just ignore the false rows but they'll be present to make things easier to update. 👍

backend/tasks/create-date-range-annual/create-date-range-annual.js
duncan-oxd: Oh, I didn't notice the name is Operating for the park level and _Operation_ for the other levels. Wow 😳 

Even so, should we add "parkLevel = true" to the `where` on that query, just in case there's a future name change that makes it the same as the other levels?

ayumi-oxd: Sounds good. Operating/Operation are basically the same, and v1 used "Operation" but the new list came with "Operating" ... we can change it in the future for consistency, so adding a level is a good idea 👍

duncan-oxd (approved):
Nice!

PR #222, opened 2025-06-20
CMS-909: Season form

duncan-oxd:
### Jira Ticket

CMS-909

### Description

Hooking up the season edit form(s) to the API data. This branch will add the ability to:

- [ ] Edit existing dates
- [ ] Add new dates
- [ ] Remove existing dates
- [ ] Save as a draft
- [ ] Submit for review (for Submitters)
- [ ] Approve (for Approvers)


frontend/src/router/index.jsx
```diff
@@ -45,76 +37,6 @@ const RouterConfig = createBrowserRouter(
             },
           ],
         },
-
-        // View park details
-        {
-          path: paths.park(":parkId"),
-          element: <ParkDetails />,
-        },
-
```
duncan-oxd, line 54: Removed all of the old pages in this ticket since they're no longer working, and all of the "TODO" comments in them are irrelevant 👍

frontend/src/components/FormPanel.jsx
```diff
@@ -224,140 +255,87 @@ function FormPanel({ show, setShow, formData, approver }) {
...
+          notes={notes}
+          setNotes={setNotes}
+          previousNotes={season.changeLogs}
+        />
+        <Buttons
+          approver={approver}
+          onApprove={onApprove}
+          onSave={() => onSave(false)}
+          onSubmit={onSubmit}
+          loading={sendingApprove || sendingSubmit || sendingSave}
```
duncan-oxd, line 303: The loading status here is very rudimentary right now. In the future, we'll probably show spinners on individual buttons and selectively enable/disable buttons while things are loading... but that's out of scope 😎

frontend/src/components/FormPanel.jsx
```diff
@@ -134,85 +49,201 @@ function Buttons({ onSave, onSubmit, approver }) {
...
+  // Clears and re-fetches the data
+  function resetData() {
+    setData(null);
+
+    // Refresh the data from the API
+    refreshData();
+  }
+
+  async function onSave(close = true) {
+    // Format the data for the API
```
ayumi-oxd, line 139: Some error handling would be nice in case, but it can be addressed in another ticket 👍

frontend/src/components/GateForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+GateForm.propTypes = {
+  gateTitle: PropTypes.string,
+  gateDescription: PropTypes.string,
+  hasGate: PropTypes.bool,
+  setHasGate: PropTypes.func.isRequired,
+  dateRanges: PropTypes.object,
+  level: PropTypes.string,
+  currentYear: PropTypes.number,
+  lastYear: PropTypes.number,
+};
```
ayumi-oxd, line 114: I'll work on here after this PR is merged 🙂

frontend/src/components/SeasonForms/AreaSeasonForm.jsx
```diff
@@ -0,0 +1,455 @@
...
+  previousSeasonDates,
+  areaDateTypes,
+  featureDateTypes,
+}) {
+  const { ROLES, checkAccess } = useAccess();
+  const approver = useMemo(
+    () => checkAccess(ROLES.APPROVER),
+    [checkAccess, ROLES.APPROVER],
+  );
+
```
ayumi-oxd, line 34: Is it possible to pass this data as a prop from the parent component `FormPanel`?

ayumi-oxd (commented):
Congrats on finishing up your big PR! I have nothing but respect that you made this 👏 It looks great 👍 I left some comments, but most of them are dups 🙂

Some other comments:
- I don't see Tier 1, Tier 2, and Winter fee date fields at the Park level, even if a park is applicable (e.g. a park has `hasTierDates` TRUE).
- I believe no `Operating dates` is needed at the Park level since it's for the gate

Screenshot 2025-07-03 at 3 33 11 PM

backend/routes/api/seasons.js
```diff
@@ -240,9 +252,42 @@ router.get(
...
+    const dateTypesByName = _.keyBy(dateTypesArray, "name");
 
-    const output = { current: seasonModel, previous: previousSeason };
+    // Return the DateTypes in a specific order
+    const orderedDateTypes = [
+      dateTypesByName.Operation,
+      dateTypesByName.Reservation,
+      dateTypesByName["Backcountry registration"],
+    ];
+
```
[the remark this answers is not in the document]
duncan-oxd: Agreed! I've been building a little list of refactoring ideas so we can add that 👍

frontend/src/components/DateRangeFields.jsx
```diff
@@ -0,0 +1,203 @@
...
+import DatePicker from "react-datepicker";
+import Form from "react-bootstrap/Form";
+import PropTypes from "prop-types";
+
+import { normalizeToUTCDate, normalizeToLocalDate } from "@/lib/utils";
+
+function DateRange({ dateRange, updateDateRange, removeDateRange }) {
+  // A unique ID for template loops and selectors
+  const idOrTempId = dateRange.id || dateRange.tempId;
+
```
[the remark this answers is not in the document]
duncan-oxd: In the "addDateRange" functions in each of the SeasonForm components, we use the `randomUUID` function to generate a temporary ID for new date ranges, so we can uniquely identify them before they get an ID from the database. V1 of the app used the same technique and it saved a lot of headaches 👍 

```js
  // Adds a new date range to the Park's dateable.dateRanges
  function addDateRange(dateType) {
    const newDateRange = {
      // Add a temporary ID for records that haven't been saved yet
      tempId: crypto.randomUUID(), // ⬅️ from the built-in crypto JS library
      startDate: null,
```

The only downside is that UUIDs are strings and our DB's IDs are numbers, but they're two different fields so it's not too confusing with the two different data types.

backend/routes/api/seasons.js
```diff
@@ -378,15 +524,27 @@ router.post(
...
+    let { readyToPublish } = req.body;
+
+    // If the user isn't an approver, they shouldn't be able to set readyToPublish
+    const isApprover = checkUserRoles(req.auth, ["doot-approver"]);
+
+    if (!isApprover) {
+      // Clear the value from the request body
+      // This will prevent the user from changing readyToPublish
+      readyToPublish = null;
+    }
```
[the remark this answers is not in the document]
duncan-oxd: All of our other role/permission checks completely block the endpoint if you don't have permission with route middleware. This was the first/only place we just wanted to check in the middle of an endpoint function so I made a very simple function to check the user's roles. Lots of room for improvement here 🥵 but it works for now

frontend/src/components/DateRangeForm.jsx
```diff
@@ -1,105 +1,23 @@
...
-  hasTier1Dates,
-  hasTier2Dates,
-  hasWinterFeeDates,
+  hasGateDates = false,
 }) {
   // Constants
   // check if there are any date ranges
-  const hasDateRanges = dateRanges && Object.keys(dateRanges).length > 0;
+  // const hasDateRanges = dateRanges && Object.keys(dateRanges).length > 0;
+  const hasDateRanges = false;
```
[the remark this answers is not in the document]
duncan-oxd: I think I just disabled a lot of stuff when I started working on the branch - the `hasDateRanges` var was just there from when you built the frontend. I think this whole `DateRangeForm` file isn't really being used in this branch anymore because the contents got split out into the 3 level SeasonForm files. I can undo this change or delete the file if we don't need it anymore. It's still included in the GateForm file in this branch so I just left it. Do we still need it? Or will GateForm use the `dateRangeFields` directly or something?

frontend/src/components/FormPanel.jsx
```diff
@@ -368,5 +346,5 @@ FormPanel.propTypes = {
   show: PropTypes.bool.isRequired,
   setShow: PropTypes.func.isRequired,
   formData: PropTypes.object,
-  approver: PropTypes.bool,
```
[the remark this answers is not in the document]
duncan-oxd: I took this out when i was splitting all of the files up for the 3 levels, and now all 3 SeasonForm files are just using the `useAccess` hook to check for the approver role. I was going to put it back here, but I figured using the hook inside the component is fine too 🤷

frontend/src/components/SeasonForms/AreaSeasonForm.jsx
duncan-oxd: I moved it to make splitting the files up easier, and then I wasn't sure if it was worth the tiny effort to put it back in FormPanel because it's not needed in that file, but it would cut down on repeated code in the child components. I'll move it back up to the parent component so we can compute it once and pass it down 👍

duncan-oxd: Oh! I see it's still being computed in FormPanel for the buttons so I don't even have to add it back there. 😅 Even better. Pushing the update now ✅

duncan-oxd:
Thanks! That's all great feedback. If you don't see editable seasons available, try running the create-seasons script I mentioned in this comment: https://github.com/bcgov/bcparks-staff-portal/pull/222#discussion_r2184172358

The create-seasons script (currently) doesn't create blank dates, just the seasons to add dates to. So I'd expect it to show nothing, but then you can click "Add more Winter fee dates" and the date inputs will appear.

The one exception is for Tier 2 dates: the script _does_ create blank dates for Tier 2 dates, only if the previous season had Tier 2 dates... And we don't have any previous seasons so that probably won't happen. 

If we fill out some 2025 Tier 2 dates at the park level, then run create-seasons fro 2026, it should have some blank dateRanges pre-created in that case.

ayumi-oxd (approved):
Thanks for adding fixes! Looks good overall 👍

PR #225, opened 2025-06-27
CMS-896: Allow clicking outside of form panel to close it

ayumi-oxd:
### Jira Ticket

CMS-896

### Description

- Allow clicking outside of the form panel to close it
- Note: Not urgent but it'd be good to come back here again to discuss whether we should display an alert/warning before closing it to avoid unintentional behaviour


duncan-oxd (approved):
There's a ticket coming up to add a bunch of warnings for scenarios like that: leaving a form without saving changes, etc. So this will be handled 👍

PR #226, opened 2025-06-28
CMS-872: Add is date range annual to park form and feature form

ayumi-oxd:
### Jira Ticket

CMS-872

### Description

- FE part of https://github.com/bcgov/bcparks-staff-portal/pull/220
- Add `DateRangeAnnuals` to `Season` based on its `publishableId`
- Add `DateRangeAnnuals` to each level of the Season API
- Display the "Dates are the same every year" checkbox checked if its `DateRangeAnnual` has `isDateRangeAnnual=TRUE`
- Disable the date picker field if `isDateRangeAnnual` is TRUE
- Update data by `onChange` event


frontend/src/components/DateRangeFields.jsx
```diff
@@ -135,16 +137,42 @@ DateRange.propTypes = {
...
+  hasMultipleDates = true,
 }) {
+  // Functions
+
+  // find the matching dateRangeAnnual for this dateType
+  const matchedDateRangeAnnual = useMemo(() => {
+    if (!dateType || !dateRangeAnnuals) return null;
+
+    return dateRangeAnnuals.find(
+      (dateRangeAnnual) => dateRangeAnnual.dateType?.name === dateType.name,
```
duncan-oxd, line 160: if `dateRangeAnnual.dateType` is undefined here, this `find` operation will return undefined - is there a chance that could ever happen? If so, maybe we should catch that and return `null` for that too or something?

(Or if it won't happen, we don't need that `?` after `dateType`)

frontend/src/components/DateRangeFields.jsx
```diff
@@ -168,12 +197,13 @@ export default function DateRangeFields({
         </button>
       )}
 
-      {/* TODO: CMS-872 - use isDateRangeAnnual */}
       <Form.Check
         type="checkbox"
-        id="same-dates-every-year"
-        name="sameDatesEveryYear"
+        id={`date-range-annual-${dateType.id}`}
+        name="DateRangeAnnual"
```
duncan-oxd, line 203: Maybe the name should be unique too (same as the `id` attribute?)`
We won't really be using the "formData" but checkboxes are meant to have unique names (and/or a `value`?)

duncan-oxd (changes requested):
Looks good! A few notes and questions; nothing major 👍

frontend/src/components/DateRangeFields.jsx
ayumi-oxd: Good catch! It's supposed to have `dateType` always, so I removed `?`.

DateRangeAnnual model:
```js
dateTypeId: {
  type: DataTypes.INTEGER,
  allowNull: false,
  ...
  },
},
```

duncan-oxd (approved):
This looks great! Please wait until #222 is merged, and then we can rebase this and merge it today 👍

PR #231, opened 2025-07-07
CMS-872: Save is date range annual update

ayumi-oxd:
### Jira Ticket

CMS-872

### Description

- Update `isDateRangeAnnual` state when the save button is clicked

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -539,7 +539,11 @@ router.post(
   sanitizePayload,
   asyncHandler(async (req, res) => {
     const seasonId = Number(req.params.seasonId);
-    const { notes = "", deletedDateRangeIds = [] } = req.body;
+    const {
+      notes = "",
+      deletedDateRangeIds = [],
+      dateRangeAnnuals = [],
+    } = req.body;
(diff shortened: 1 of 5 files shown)


backend/routes/api/seasons.js
```diff
@@ -571,6 +575,24 @@ router.post(
       // If readyToPublish is null or undefined, set it to the current value
       const newReadyToPublish = readyToPublish ?? season.readyToPublish;
 
+      // dateRangeAnnuals
+      const dateRangeAnnualsToSave = (dateRangeAnnuals || [])
+        .filter((dateRangeAnnual) => dateRangeAnnual.changed)
```
duncan-oxd, line 580: With the date ranges, we do this filtering on the frontend so we only send the changed stuff to the APi. Would it be possible to adjust this to use the same pattern? Then the API endpoint can just send whatever is sent to it, and there's less stuff sent over the network.

duncan-oxd (changes requested):
Nice! One small change request to keep the pattern used for the rest of the API stuff when we save.

duncan-oxd (approved):

PR #233, opened 2025-07-08
CMS-1024: Hide dates are the same every year checkbox for tier 1 and tier 2

ayumi-oxd:
### Jira Ticket

CMS-1024
CMS-1033

### Description

- CMS-1024: Hide the dates are the same every year checkbox for tier 1 and tier 2
- CMS-1033: Fix a spacing issue

### DIFF
--- a/backend/tasks/create-date-range-annual/create-date-range-annual.js
+++ b/backend/tasks/create-date-range-annual/create-date-range-annual.js
@@ -88,15 +88,10 @@ export async function createDateRangeAnnualEntries() {
 
           if (!dateTypeId) continue;
 
-          // skip if dateType name is "Tier 1" or "Tier 2" or "Operating"
+          // skip if dateType name is "Tier 1" or "Tier 2"
           const dateTypeName = dateTypeNameById[dateTypeId];
 
-          if (
-            dateTypeName === "Tier 1" ||
-            dateTypeName === "Tier 2" ||
-            dateTypeName === "Operating"
-          )
-            continue;
+          if (dateTypeName === "Tier 1" || dateTypeName === "Tier 2") continue;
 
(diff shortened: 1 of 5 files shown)


backend/tasks/create-date-range-annual/create-date-range-annual.js
```diff
@@ -88,15 +88,10 @@ export async function createDateRangeAnnualEntries() {
...
+          // skip if dateType name is "Tier 1" or "Tier 2"
           const dateTypeName = dateTypeNameById[dateTypeId];
 
-          if (
-            dateTypeName === "Tier 1" ||
-            dateTypeName === "Tier 2" ||
-            dateTypeName === "Operating"
-          )
-            continue;
+          if (dateTypeName === "Tier 1" || dateTypeName === "Tier 2") continue;
```
ayumi-oxd, line 95: That was my mistake in the previous PR 😅 Operating dates can have a DateRangeAnnual, but not Tier 1 and Tier 2.

frontend/src/components/DateRangeFields.jsx
```diff
@@ -189,22 +194,26 @@ export default function DateRangeFields({
...
+          type="checkbox"
+          id={`date-range-annual-${dateType.id}`}
+          name={`date-range-annual-${dateType.id}`}
+          label="Dates are the same every year"
+          checked={isDateRangeAnnual}
+          onChange={handleDateRangeAnnualChange}
+          className="mt-2 mb-0"
+        />
+      )}
     </>
```
ayumi-oxd, line 217: From the script, DateRangeAnnual was not supposed to be created if the date type is Tier 1/Tier 2. However, it's possible to add those from AdminJS, so I added another condition in the frontend.

duncan-oxd (approved):

PR #235, opened 2025-07-09
CMS-1046: Upgrade CrunchyDB to Postgres 17

molund:
### Jira Ticket

CMS-1046

### Description
Updated CrunchyDB to Postgres 17 based on a request from OCIO.  Alpha-dev and Alpha-test are techincally still on 16 because I tried to update them in 2 steps and I was unsuccessful at updating from 16-17.

### DIFF
--- a/helm/crunchy-postgres/Upgrade-Notes.md
+++ b/helm/crunchy-postgres/Upgrade-Notes.md
@@ -1,0 +1,54 @@
+# This is based on these instructions
+
+https://access.crunchydata.com/documentation/postgres-operator/5.7/guides/major-postgres-version-upgrade
+
+# Create PGUpgrade.yml file
+
+```
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PGUpgrade
+metadata:
+  name: crunchy-postgres-upgrade
+spec:
+  postgresClusterName: crunchy-postgres
+  fromPostgresVersion: 15
+  toPostgresVersion: 17
+```
+
+# Create the resource on OpenShit
+
+```
+oc create -n a7dd13-prod -f PGUpgrade.yml
+```
+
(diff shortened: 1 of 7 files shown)


helm/crunchy-postgres/Upgrade-Notes.md
```diff
@@ -0,0 +1,54 @@
...
+
+```
+oc create -n a7dd13-prod -f PGUpgrade.yml
+```
+
+# Annotate the postgrescluster to allow upgrade and patch it to shutdown
+
+```
+kubectl -n a7dd13-prod annotate postgrescluster crunchy-postgres postgres-operator.crunchydata.com/allow-upgrade="crunchy-postgres-upgrade"
+kubectl -n a7dd13-prod patch postgrescluster crunchy-postgres --type=merge -p '{"spec":{"shutdown":true}}'
```
duncan-oxd, line 28: It probably doesn't make a difference, but should we use `oc` instead of `kubectl` in these lines?

molund: I got that recommendation from co-pilot too, but I kept it the way it was to match the article I used for instructions.

molund:
> That's awesome 🙌 For the dev environment, the db dev container is defined with this:
> 
> ```dockerfile
>   db:
>     image: postgres:15
> ```
> 
> I imagine we can just change that from 15 to 17 and then rebuild the container with no other changes needed, but I don't want to try it in the middle of this sprint! If you've got a fairly fresh local dev environment, are you able to try that and see if it works? If so, we can include the change in this branch and then I'll just avoid rebuilding until I'm between sprints.

I tested it. Everyting seems to work okay with 17 when you do a fresh install, but there isn't a nice upgrade path for local dev.

PR #236, opened 2025-07-09
CMS-1043: Copy previous year dates value if dates are the same every year

ayumi-oxd:
### Jira Ticket

CMS-1043

### Description

- Copy previous year dates (if it has) to the current year dates if `isDateRangeAnnual` is TRUE
- Add previous year dates (if it has) to the form if a user clicks the Add more date button when `isDateRangeAnnual` is TRUE
- Adjust some spacing/layout of the gate form
- TODO: Add script to add DateRanges with the previous year dates if a Season has DateRangeAnnual but no DateRange?

### DIFF
--- a/frontend/src/components/DateRangeFields.jsx
+++ b/frontend/src/components/DateRangeFields.jsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from "react";
+import { useMemo, useState, useEffect } from "react";
 import {
   faPlus,
   faXmark,
@@ -13,6 +13,7 @@ import { normalizeToUTCDate, normalizeToLocalDate } from "@/lib/utils";
 
 function DateRange({
   dateRange,
+  previousDateRanges = [],
   updateDateRange,
   removeDateRange,
   isDateRangeAnnual,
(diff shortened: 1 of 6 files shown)


ayumi-oxd:
@duncan-oxd It was more complicated than I thought, so I'm not 100% sure if this was the best way. Any advice would be appreciated!

duncan-oxd:
I think (based on the ticket description) we may just want to do this one time, in advance, when the new requested season is created in the db, and that happens in the backend create-seasons.js script. 
If we do it there, then we can set eveything up according to the requirements, and then the user can uncheck the box or edit the dates if they want to, but then we don't need to handle any special logic on the frontend. How does that sound?

ayumi-oxd:
@duncan-oxd Please ignore this for now since I created another PR: https://github.com/bcgov/bcparks-staff-portal/pull/239

duncan-oxd:
Should we label this one or close it if the other PR will take its place?

ayumi-oxd:
Close it as https://github.com/bcgov/bcparks-staff-portal/pull/239 was merged.

PR #238, opened 2025-07-10
CMS-1014: Implement approve button on table rows

duncan-oxd:
### Jira Ticket

CMS-1014

### Description

This hooks up the "Approve" button on the homepage table. It does the same thing as if you opened the form and clicked "Mark approved" in there:
- Update the status to "Approved"
- Reload the data in the table to show the changes
- Show a success flash message

I added yet another context to provide the main table's refreshTable function down to the ApproveButton, rather than passing it down 5 or 6 levels. 

I also renamed the "renderTable" functional component to ParksTableWrapper and used a proper tag for it in the JSX. I think that was left over from v1 when it was some of the first React code I ever wrote 🙈 

This branch is pretty quick and basic; I put in a TODO comment about handling API errors and I'll be sure to add something to the backlog for that. I also had the idea to return some season details from the API when we save/approve/submit so we will have all the info we need to show in he flash message, rather than having to gather it from the frontend. That hasn't really been a problem yet but it could be. I'll put that as an optimization/refactoring task in the backlog too.

### DIFF
--- a/frontend/src/components/EditAndReviewTable.jsx
+++ b/frontend/src/components/EditAndReviewTable.jsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useContext } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import { faCheck } from "@fa-kit/icons/classic/solid";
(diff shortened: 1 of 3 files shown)


duncan-oxd:
And I'll tag Amanda in a comment about disabling the button if it's already approved, because that seems like a good idea!

PR #239, opened 2025-07-11
CMS-1043: Populate date ranges if dates are the same every year

ayumi-oxd:
### Jira Ticket

CMS-1043

### Description

- Create this PR to try a different approach for CMS-1043, based on our conversation in https://github.com/bcgov/bcparks-staff-portal/pull/236
- Populate date ranges based on the previous year's dates if `isDateRangeAnnual` is TRUE
- Case 1: Populate dates when Seasons are newly created 
  - e.g. Run `create-season 2026`
  - If the date type has `isDateRangeAnnual` is TRUE, 2026 dates will be populated based on 2025 dates
 - Case 2: Populate dates when the current Seasons do not have dates
   - e.g. Run `populate-date-ranges 2025`
   - If the date type has `isDateRangeAnnual` is TRUE and no DateRange of the year, 2025 dates will be populated based on 2024 dates

### DIFF
--- a/backend/tasks/populate-date-ranges/README.md
+++ b/backend/tasks/populate-date-ranges/README.md
@@ -18,7 +18,7 @@ This script populates `DateRange` records for a given target year based on the p
 From your project root, run:
 
 ```sh
-node backend/tasks/populate-date-ranges/populate-date-ranges.js 2026
+node tasks/populate-date-ranges/populate-date-ranges.js 2026
 ```
 
 Replace 2026 with the year you want to populate.
--- a/backend/tasks/populate-date-ranges/README.md
+++ b/backend/tasks/populate-date-ranges/README.md
@@ -1,0 +1,41 @@
+# populate-date-ranges.js
+
+This script populates `DateRange` records for a given target year based on the previous year's `DateRange` entries, but **only if the associated `DateRangeAnnual` is marked as annual** (`isDateRangeAnnual` is `TRUE`). It copies the month and day from the previous year's dates, but updates the year to the target year.
+
+## What does the script do?
+
+1. **For each DateRangeAnnual marked as annual:**
+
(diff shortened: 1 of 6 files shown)


backend/tasks/populate-date-ranges/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+2. **Transaction Safety:**
+   - All operations are performed inside a transaction. If any error occurs, all changes are rolled back.
+
+## How to run
+
+From your project root, run:
+
+```sh
+node backend/tasks/populate-date-ranges/populate-date-ranges.js 2026
```
duncan-oxd, line 21: I just thought of something while reviewing #237 and it applies here too:

> This should be run in the backend dev container (so the appropriate environment variables will all be set) so we won't have to specify the backend/ part here. (it probably works because of the env.js file, but it's best to run it from the backend container)

backend/tasks/populate-date-ranges/populate-date-ranges.js
```diff
@@ -0,0 +1,138 @@
...
+      console.log("No new DateRanges to create.");
+    }
+
+    await transaction.commit();
+    console.log(
+      "DateRanges populated for new Seasons based on previous year's annual DateRanges.",
+    );
+  } catch (err) {
+    await transaction.rollback();
+    console.error("Error populating DateRanges from annual:", err);
```
duncan-oxd, line 118: Since the transaction comes from a parameter now (and it could be null), let's move the commit/rollback code out of here too. The parent will decide what to do with the transaction.

In this file, we're creating a transaction on line 125 if it's run directly, so we can commit or rollback in that same section 👍

duncan-oxd (changes requested):
One more thing to move, and the rest looks good!

duncan-oxd (approved):

PR #247, opened 2025-07-15
CMS-1081: Update ready to publish copy

ayumi-oxd:
### Jira Ticket

CMS-1081

### Description

- Update ready to publish copy
- Change the toggle from On/Off to Yes/No


duncan-oxd (approved):
I'd still pick a checkbox 10 times out of 10 instead of one of these apple-style toggle switches 😂 But maybe I'm old fashioned

PR #248, opened 2025-07-15
CMS-1070: Omit Area-level dates from Area form

duncan-oxd:
### Jira Ticket

CMS-1070

### Description

The Area Season form won't show any area-level dates. Every Area has at least one feature (even if it's a fake feature we created called "All sites"). I thought we'd use the area-level to create dates that applied to all the features, but it sounds like we're not doing that at the moment.

The area-level form will still show feature dates, and the gate hours will be the only area-level thing on it. Let me know if my changes will cause problems with the gates section, but I think it won't affect it!

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -400,25 +400,18 @@ router.get(
       [Op.or]: [{ parkAreaLevel: true }, { featureLevel: true }],
     });
 
-    // Include all DateTypes for the Park Area level
-    const areaDateTypesArray = await getAllDateTypes({
-      parkAreaLevel: true,
-    });
-
     // Include all DateTypes for the Feature level
     const featureDateTypesArray = await getAllDateTypes({
       featureLevel: true,
     });
 
-    const areaDateTypesByName = _.keyBy(areaDateTypesArray, "name");
     const featureDateTypesByName = _.keyBy(featureDateTypesArray, "name");
 
     // Return the DateTypes in a specific order
-    const orderedAreaDateTypes = [
-      areaDateTypesByName.Operation,
-      areaDateTypesByName.Reservation,
-      areaDateTypesByName["Backcountry registration"],
-    ];
+
+    // Don't include any Area-level dates.
+    // Area forms will only have Feature-level dates.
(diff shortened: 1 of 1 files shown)


backend/routes/api/seasons.js
```diff
@@ -453,7 +441,9 @@ router.get(
     const output = {
       current: currentSeason,
       previous: previousSeason,
-      areaDateTypes: orderedAreaDateTypes,
+      // Don't include any Area-level dates.
+      // Area forms will only have Feature-level dates.
+      areaDateTypes: [],
       featureDateTypes: orderedFeatureDateTypes,
```
ayumi-oxd, line 447: What is `areaDateTypes` for?

ayumi-oxd (commented):
Looks good! Just have some questions 🙂

Wondering if we might need to remove the date fields from `AreaSeasonForm`. 

```js
{areaDateTypes.map((dateType) => (
  <div key={dateType.name} className="col-lg-6 mb-4">
    <h6 className="fw-normal">
      {dateType.name}{" "}
      <TooltipWrapper placement="top" content={dateType.description}>
        <FontAwesomeIcon icon={faCircleInfo} />
      </TooltipWrapper>
    </h6>

    <PreviousDates
      dateRanges={previousAreaDatesByType?.[dateType.name]}
    />

    <DateRangeFields
      dateableId={parkArea.dateableId}
      dateType={dateType}
      dateRanges={areaDatesByType[dateType.name] ?? []}
      updateDateRange={updateAreaDateRange}
      addDateRange={addAreaDateRange}
      removeDateRange={removeAreaDateRange}
      dateRangeAnnuals={dateRangeAnnuals}
      updateDateRangeAnnual={updateDateRangeAnnual}
    />
  </div>
))}
```

backend/routes/api/seasons.js
ayumi-oxd: Makes sense to me 👍 So we'll need to update the DateTypes too.

ayumi-oxd (approved):
Looks good! Let's just remember that we probably want to revisit here in the refactoring phase 🙂

PR #249, opened 2025-07-16
CMS-1075: Update park endpoint

ayumi-oxd:
### Jira Ticket

CMS-1075

### Description

- Update the park endpoint
- Remove `description` from `DateType` since it's not necessary for the table
- `Feature` in `ParkArea` didn't have a proper `Season` at the park endpoint since `ParkArea.Feature` is not publishable. When a new date is added to `ParkArea.Feature`, it goes to `ParkArea.Season`, not `ParkArea.Feature.Season`. To fix this, I updated `buildFeatureOutput` to get the related `DateRange` by `dateableId` from `ParkArea.Season`

### DIFF
--- a/frontend/src/components/EditAndReviewTable.jsx
+++ b/frontend/src/components/EditAndReviewTable.jsx
@@ -183,8 +183,6 @@ function Table({ park, formPanelHandler }) {
   const parkAreas = park.parkAreas || [];
   const features = park.features || [];
 
-  console.log("park:", park);
-
   return (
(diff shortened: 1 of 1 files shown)


duncan-oxd (approved):
Looks good 👍  The queries for /season endpoints are pretty detailed and they get lots of the same info, so we could probably figure out a way to use the same (or similar) query and formatting in both places some time. When we get to a refactoring phase!

PR #251, opened 2025-07-16
CMS-907: Add gate detail to season change log

ayumi-oxd:
### Jira Ticket

CMS-907

### Description

- Track `GateDetail` change in `SeasonChangeLog`
- Remove `gateOpen24Hours` from the `getGateDetail` function since it's not needed for now
- For testing
  - Change something related to `GateDetail` and save it, then check the `SeasonChangeLog` on AdminJS
  - `GateDetail` change will appear in a JSON object format

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -49,15 +49,6 @@ function checkSeasonExists(season) {
 }
 
 /**
- * Returns a plain JS object from a Sequelize instance or plain object.
- * @param {Object} obj Sequelize instance or plain object
- * @returns {Object} Plain JS object
- */
-function convertToPlainObject(obj) {
-  return obj && typeof obj.toJSON === "function" ? obj.toJSON() : obj;
-}
-
-/**
  * Updates the status of a Season.
  * If "save" is true, it will first save the changes to the Season.
  * @param {number} seasonId The ID of the season to update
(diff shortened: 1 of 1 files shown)


backend/migrations/20250716043055-add-gate-detail-to-season-change-log.js
```diff
@@ -0,0 +1,20 @@
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn("SeasonChangeLogs", "gateDetailOldValue", {
+      type: Sequelize.TEXT,
```
duncan-oxd, line 5: If we use a JSONB column type, we can put json data in here without having to stringify it:
```js
Sequelize.JSONB
```

That's what we did for the managementAreas column in the Parks table, for an example 👍

duncan-oxd (changes requested):
The changes look good, but we can simplify things by using a JSONB column instead of storing it as text. I put a suggestion in a comment for the migration. Let me know if that sounds good or if you have any questions 👍

backend/migrations/20250716043055-add-gate-detail-to-season-change-log.js
ayumi-oxd: Good to know about `JSONB`, thanks! Updated it.

backend/routes/api/seasons.js
```diff
@@ -48,6 +48,15 @@ function checkSeasonExists(season) {
   throw error;
 }
 
+/**
+ * Returns a plain JS object from a Sequelize instance or plain object.
+ * @param {Object} obj Sequelize instance or plain object
+ * @returns {Object} Plain JS object
+ */
+function convertToPlainObject(obj) {
+  return obj && typeof obj.toJSON === "function" ? obj.toJSON() : obj;
```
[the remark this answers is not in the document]
ayumi-oxd: Yes, you're right! It works without converting it 👍

PR #252, opened 2025-07-17
CMS-898: Temporarily disable winter fee propagation to features

duncan-oxd:
### Jira Ticket

CMS-898

### Description

After discussing this feature with UX and QA today, we realized it's not needed until later on, and the logic may change. Amanda asked to simply disable it for now, and hide Winter fees at the feature/area level on the homepage, and we'll revisit it in a future sprint.

(We're not collecting winter fees at those levels, they'd be generated by the `propagateWinterFeeDates` script which is being commented out here, but the temporary `filter` lines will hide any that exist in the DB for any other reason)

### DIFF
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -158,7 +158,10 @@ function buildFeatureOutput(
     ),
   );
   // get date ranges for park.feature
-  const featureDateRanges = getAllDateRanges(filteredSeasons);
+  const featureDateRanges = getAllDateRanges(filteredSeasons)
+    // Temporarily disabling display of Winter Fees
+    // @TODO: Remove this filter when Winter fee logic is revised (CMS-898)
+    .filter((dateRange) => dateRange.dateType?.name !== "Winter fee");
 
   const output = {
     id: feature.id,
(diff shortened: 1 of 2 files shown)


backend/routes/api/seasons.js
```diff
@@ -28,7 +28,7 @@ import {
 } from "../../middleware/permissions.js";
 
 // import { createFirstComeFirstServedDateRange } from "../../utils/firstComeFirstServedHelper.js";
```
duncan-oxd, line 30: I noticed the FCFS script is still commented out 😬 , I disabled it temporarily to avoid breaking things while the edit forms were being developed. Let's chat about this, but do we need to re-enable it in a ticket somewhere, or is it okay to still be disabled? Sorry for not mentioning that sooner!

backend/routes/api/parks.js
```diff
@@ -148,7 +148,10 @@ function getAllDateRanges(seasons) {
 // build feature output object
 function buildFeatureOutput(feature, currentYear, includeCurrentSeason = true) {
   // get date ranges for park.feature
-  const featureDateRanges = getAllDateRanges(feature.seasons);
+  const featureDateRanges = getAllDateRanges(feature.seasons)
+    // Temporarily disabling display of Winter Fees
+    // @TODO: Remove this filter when Winter fee logic is revised (CMS-898)
+    .filter((dateRange) => dateRange.dateType?.name !== "Winter fee");
```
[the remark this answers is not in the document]
duncan-oxd: I won't bother for this temporary change

backend/routes/api/seasons.js
ayumi-oxd: Good catch! I'll work on it again before CMS-957 but after CMS-1014 since the FCFS needs an approval function. So I need to review your PR first 😅

PR #253, opened 2025-07-17
CMS-1002: Hide form sections with showDateFormSections

duncan-oxd:
### Jira Ticket

CMS-1002

### Description

When there are no applicable date types to collect, the "BC Parks Reservations" box can sometimes be empty, and in that case we don't want to show it. This ticket adds some "show ___ section" variables and wraps the template sections to prevent them from rendering if they'll be empty. This is somewhat related to the other 4 or 5 branches we just merged involving this BC Parks Reservations box. This scenario is getting increasingly unlikely as the other fixes have been merged, but it's still possible with certain conditions, so this branch prevents empty form sections from rendering.

In the Park and Feature forms, it's just one item, so the dates will either be all in the box, or all outside of the box. Just one variable `showDateFormSections` decides whether to show or hide everything.

In the Area form, you could theoretically have some features in the BCP Res. section and some that aren't, so this form uses two variables: `showBcpResSection` and `showNonBcpResSection` to show or hide the two sections individually, but the pattern is pretty much the same.

### DIFF
--- a/frontend/src/components/SeasonForms/AreaSeasonForm.jsx
+++ b/frontend/src/components/SeasonForms/AreaSeasonForm.jsx
@@ -287,6 +287,18 @@ export default function AreaSeasonForm({
     [features],
   );
 
+  // Show the BC Parks Reservations section if there are any area date types,
+  //  or any features with inReservationSystem
+  const showBcpResSection = useMemo(
+    () => areaDateTypes.length > 0 || bcpResFeatures.length > 0,
+    [areaDateTypes.length, bcpResFeatures.length],
+  );
+
+  const showNonBcpResSection = useMemo(
+    () => nonBcpResFeatures.length > 0,
+    [nonBcpResFeatures.length],
+  );
+
   const featureDatesByType = useMemo(() => {
     // Create objects with IDs as keys to loop over
     const featuresByDateableId = keyBy(features, "dateableId");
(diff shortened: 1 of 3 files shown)


frontend/src/components/SeasonForms/AreaSeasonForm.jsx
```diff
@@ -287,6 +287,18 @@ export default function AreaSeasonForm({
     [features],
   );
 
+  // Show the BC Parks Reservations section if there are any area date types,
+  //  or any features with inReservationSystem
+  const showBcpResSection = useMemo(
+    () => areaDateTypes.length > 0 || bcpResFeatures.length > 0,
+    [areaDateTypes.length, bcpResFeatures.length],
+  );
+
```
ayumi-oxd, line 296: I guess `areaDateTypes` always has a length of 0 as it's an empty array. Does it need to be here?

frontend/src/components/SeasonForms/AreaSeasonForm.jsx
```diff
@@ -436,40 +448,66 @@ export default function AreaSeasonForm({
...
-                addDateRange={addAreaDateRange}
-                removeDateRange={removeAreaDateRange}
-                dateRangeAnnuals={dateRangeAnnuals}
-                updateDateRangeAnnual={updateDateRangeAnnual}
+      {showBcpResSection && (
+        <FormContainer>
+          <div className="row">
+            {/* Area-level dates */}
+            {areaDateTypes.map((dateType) => (
+              <div key={dateType.name} className="col-lg-6 mb-4">
```
ayumi-oxd, line 456: I believe it won't be rendered, but would you like to keep it here? Since we don't create any dates at the Park Area level.

ayumi-oxd (commented):
Looks good! Just have some q's. I might have missed the recent changes so I just might be wrong 😅

frontend/src/components/SeasonForms/AreaSeasonForm.jsx
duncan-oxd: Same as in the other comment, in this version it'll keep everything backwards compatible (and working as described in the requirements for CMS-909) but the API will send an empty array so it will be hidden. If the API starts sending dates, then the frontend will show them according to the requirements. That feels like a good tradeoff at the moment.

PR #257, opened 2025-07-18
CMS-1075: Remove dates row at park area level from table

ayumi-oxd:
### Jira Ticket

CMS-1075

### Description

- Remove the dates row at the park area level from the table
- `ParkArea` doesn't have dates anymore


duncan-oxd (approved):
I see why you asked those questions in #253 😆 Maybe this change is more permanent than I was thinking! I think a discussion needs to be had about what a Park Area even is, and what they hope to use it for.

duncan-oxd:
Looks good! I'll merge this so we can include it in the demo 👍

PR #260, opened 2025-07-18
CMS-1043: Add dateable id to date range

ayumi-oxd:
### Jira Ticket

CMS-1043

### Description

- Additional change on top of https://github.com/bcgov/bcparks-staff-portal/pull/239
- Some dates are displayed in the table but not in the form, because some `DateRanges` were missing `dateableId` 
- Update scripts to add `dateableId` when `DateRanges` are created
  - populate-date-ranges.js: When `Seasons` are created, `DateRanges` will be created if `dateRangeAnnual` is TRUE
  - populate-park-gate-dates.js: When the operating dates are imported from Strapi, `DateRanges` will be created
 - Add a new script to add `dateableId` to `DateRanges` if it is missing
   - For testing: 
     - Check `DateRange` on AdminJS, see the entries without `dateableId`
     - Run `node tasks/populate-missing-dateable-id/populate-missing-dateable-id.js`
     - Check `DateRange` again to see if `dateableId` is added

### DIFF
--- a/backend/tasks/populate-date-ranges/populate-date-ranges.js
+++ b/backend/tasks/populate-date-ranges/populate-date-ranges.js
@@ -4,6 +4,7 @@
 import "../../env.js";
 
 import { Season, DateRange, DateRangeAnnual } from "../../models/index.js";
+import { findDateableIdByPublishableId } from "../../utils/findDateableIdByPublishableId.js";
 
 // Functions
 /**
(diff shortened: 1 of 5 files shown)


ayumi-oxd:
> Looks good! Am I correct to think this is something we'll only need to run one time? And probably not at all on production now that the populate-date/gate-ranges script is updated?
> 
> Good idea to have a helper function to find the dateable from the publishable. 👍

Yes that's correct. If the populate-date/gate-ranges script works as expected, no need to run the populate-missing-dateable-id on DEV/TEST/PROD. I'll add it to the post-deployment ticket.

PR #263, opened 2025-07-23
CMS-1119: Remove operating dates from table if park has hasGate false

ayumi-oxd:
### Jira Ticket

CMS-1119

### Description

- Clear all gate-related data if it's been changed to `hasGate` False at Park/Area/Feature level
- It retains the data until the form is saved
- Hide operating dates from the table if a park has `hasGate` FALSE

### DIFF
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -313,8 +313,16 @@ router.get(
       ],
     });
 
+    // constrain GateDetail query to only publishableIds in parks
+    const publishableIds = parks.map((park) => park.publishableId);
+
     const allGateDetails = await GateDetail.findAll({
       attributes: ["publishableId", "hasGate"],
+      where: {
+        publishableId: {
+          [Op.in]: publishableIds,
+        },
+      },
     });
 
     const gateDetailMap = new Map();
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -17,15 +17,6 @@ import asyncHandler from "express-async-handler";
 // Constants
 const router = Router();
 
-const allGateDetails = await GateDetail.findAll({
-  attributes: ["publishableId", "hasGate"],
-});
-const gateDetailMap = {};
-
-allGateDetails.forEach(gate => {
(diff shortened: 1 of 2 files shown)


backend/routes/api/parks.js
```diff
@@ -305,9 +313,21 @@ router.get(
       ],
     });
 
+    const allGateDetails = await GateDetail.findAll({
+      attributes: ["publishableId", "hasGate"],
+    });
```
duncan-oxd, line 318: Would it be worth it to constrain this query to only the publishableIds in the parks array?

backend/routes/api/parks.js
```diff
@@ -305,9 +313,21 @@ router.get(
...
+
+    const gateDetailMap = {};
+    
+    allGateDetails.forEach((gate) => {
+      gateDetailMap[gate.publishableId] = gate.hasGate;
+    });
```
duncan-oxd, line 324: For future reference, you could use a `new Map()` here for more straightforward inserting and lookups, but either way works just fine 👍

frontend/src/components/FormPanel.jsx
```diff
@@ -188,7 +188,33 @@ function SeasonForm({
       seasonDateRanges.push(...areaDateRanges, ...featureDateRanges);
     }
 
-    const changedDateRanges = seasonDateRanges
+    // Find the "Operating" date type id
+    const operatingDateTypeId = seasonMetadata?.dateTypes?.find(
+      (dateType) => dateType.name === "Operating",
+    )?.id;
```
duncan-oxd, line 194: There are a few other spots where I needed to get the Operating type ID, so I used lodash `keyBy` on the dateTypes array, then it creates an object like `{ Operating: <operating object>, ...}`. That might be a bit cleaner to get it than doing a `find` with all the `?`s.

It _should_ always be defined, but I guess we can't guarantee that.. If it's ever not found, it will affect all of the filters below.

duncan-oxd (approved):
Looks good! I had some thoughts about syntax, but nothing that needs to change necessarily 👍

PR #264, opened 2025-07-25
CMS-1025: Hide add more dates button under tier 1 date

ayumi-oxd:
### Jira Ticket

CMS-1025

### Description

- Hide the "Add more dates" button under the Tier 1 date field
- Exception: Display the button if there's no date range

### DIFF
--- a/frontend/src/components/DateRangeFields.jsx
+++ b/frontend/src/components/DateRangeFields.jsx
@@ -108,6 +108,7 @@ export default function DateRangeFields({
   updateDateRangeAnnual,
 }) {
   // Constants
+  // Tier 1 only allows 1 date range
   const hasMultipleDates = dateType.name !== "Tier 1";
 
   // Functions


frontend/src/components/DateRangeFields.jsx
```diff
@@ -106,10 +106,11 @@ export default function DateRangeFields({
   dateType,
   dateRangeAnnuals,
   updateDateRangeAnnual,
-  hasMultipleDates = true,
 }) {
-  // Functions
+  // Constants
+  const hasMultipleDates = dateType.name !== "Tier 1";
```
duncan-oxd, line 111: Before we merge, let's just add a comment here for the future, something like "Tier 1 dates only allow 1 date range"

duncan-oxd (approved):
Looks good! I put one note in there about adding a comment but the changes are good to go 👍

PR #266, opened 2025-07-28
CMS-1089: Add error page for login

molund:
### Jira Ticket

CMS-1089

### Description
Added an error message to the V2 app for users with no keycloak roles

Note: In order to test this locally with BCeID you will need to change your keycloak to test instead of dev. The Keycloak IDP on dev has the wrong secret key and it cannot log in.

### DIFF
--- a/frontend/src/components/Unauthorized.jsx
+++ b/frontend/src/components/Unauthorized.jsx
@@ -1,0 +1,12 @@
+export function Unauthorized() {
+  return (
+    <div className="container">
+      <h1 className="mt-3 mb-2">Staff web portal</h1>
+      <p>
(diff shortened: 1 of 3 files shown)


duncan-oxd:
All of the DOOT users are in groups to inherit the group's roles:
image

What do you think about adding a general "doot-user" role to the top-level DOOT Users group, and then checking for that? All existing users would automatically inherit it, and then it would save us having to check the other roles. (I haven't tried it, but that seems like a straightforward way to allow or deny access to DOOT)

ayumi-oxd (approved):
Couldn't test it as I don't have a business BCeID, but the code looks good to me 👍

PR #269, opened 2025-07-29
CMS-1020: Update previous and current reference

ayumi-oxd:
### Jira Ticket

CMS-1020

### Description

- Update previous and current reference
- Use the most recent season as the current season
- Use the most recent -1 season as the previous season
- For the group camping Area/Feature, collect dates 2 years in advance, so the table should look like this:
 
Screenshot 2025-07-29 at 11 45 49 AM


duncan-oxd (approved):
Looks good!

PR #278, opened 2025-08-01
CMS-1082: Add "Optional" labels, disable "X" for single dates

duncan-oxd:
### Jira Ticket

CMS-1082

### Description

Added a new function to return true if a date type is "optional" at the given level. If it's optional, we show a label under the date type header. 

Also added some logic to hide and disable the "X" button to remove dates:
- If a date is not optional, you won't be able to remove them all; there must always be one left.
- If a date is optional, you can still remove them all.
- Instead of removing the button entirely, I made it disabled and invisible so the flex box elements stay the same size. We can update this with a CSS grid later on...

Added a CSS class for the "secondary' color, which is very close to black. Some things in Figma are colored in this dark grey shade so we can use this class for that. The "secondary" color in the Bootstrap is blue 🙃 so I called it "secondary-grey"

### DIFF
--- a/frontend/src/components/DateRangeFields.jsx
+++ b/frontend/src/components/DateRangeFields.jsx
@@ -1,5 +1,6 @@
 import { useMemo } from "react";
 import PropTypes from "prop-types";
+import classNames from "classnames";
 import { faPlus, faXmark } from "@fa-kit/icons/classic/regular";
 import { startOfYear, endOfYear, addYears } from "date-fns";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
@@ -12,6 +13,7 @@ function DateRange({
   dateRange,
   updateDateRange,
   removeDateRange,
+  removable = true,
   isDateRangeAnnual,
 }) {
   // A unique ID for template loops and selectors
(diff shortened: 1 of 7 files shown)


frontend/src/lib/isDateTypeOptional.js
```diff
@@ -0,0 +1,16 @@
+export const optionalTypes = {
+  // @TODO: use a list of constants for date types
```
duncan-oxd, line 2: Maybe I shouldn't leave so many "TODO" comments in my code and just capture this in a backlog ticket instead... 🤔 Any preference?

ayumi-oxd: We can create a ticket to implement "All TODOs in code" 😄

ayumi-oxd (approved):
LGTM!

frontend/src/components/DateRangeFields.jsx
```diff
@@ -148,12 +157,13 @@ export default function DateRangeFields({
...
   return (
     <>
-      {dateRanges.map((dateRange) => (
+      {dateRanges.map((dateRange, index) => (
         <DateRange
           key={dateRange.id || dateRange.tempId}
           dateRange={dateRange}
           updateDateRange={updateDateRange}
           removeDateRange={removeDateRange}
+          removable={optional || index > 0}
```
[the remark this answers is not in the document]
duncan-oxd: No, the index is the array index so there will always be one with index 0 that can't be removed.

PR #279, opened 2025-08-01
CMS-1139: Remove RSO and PO roles and add CONTRIBUTOR

molund:
### Jira Ticket

CMS-1139

### Description
I just updated the constants. I couldn't find anywhere in the code that the roles were being referenced.


duncan-oxd (approved):
That's right, as of yet no PO or RSO has ever signed into the app 😅

PR #283, opened 2025-08-08
CMS-957: Update first come first served function

ayumi-oxd:
### Jira Ticket

CMS-957

### Description

- When an approver approves the Operating dates and the Reservation dates for "Frontcountry campground" or "Walk-in camping", the FCFS dates are calculated and added to DateRanges automatically

### DIFF
--- a/backend/utils/firstComeFirstServedHelper.js
+++ b/backend/utils/firstComeFirstServedHelper.js
@@ -28,6 +28,11 @@ import {
 // First come 2: 2025-10-13 to 2025-10-30
 // => From Reservation end date plus 1 to Operation end date minus 1
 
+/**
+ * Determines if a feature type is eligible for FCFS date range generation
+ * @param {string} featureTypeName The name of the feature type to check
+ * @returns {boolean} True if the feature type supports FCFS, false otherwise
+ */
 function isFCFSFeatureType(featureTypeName) {
   if (!featureTypeName) return false;
   // Handles both Sequelize instance and plain object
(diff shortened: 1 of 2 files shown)


backend/utils/firstComeFirstServedHelper.js
```diff
@@ -21,33 +28,110 @@ import { DateRange, DateType } from "../models/index.js";
...
-
       transaction,
     });
 
     const dateTypeMap = new Map(
       dateTypes.map((dateType) => [dateType.name, dateType.id]),
     );
 
-    const operationDateRange = season.dateRanges.find(
+    // Fetch all DateRanges for this season with their DateType
```
duncan-oxd, line 124: If we're filtering by dateableId, this query doesn't get _all_ the dateRanges anymore

backend/utils/firstComeFirstServedHelper.js
```diff
@@ -21,33 +28,110 @@ import { DateRange, DateType } from "../models/index.js";
...
+      transaction,
+    });
+
+    const operationDateRange = dateRanges.find(
       (dateRange) => dateRange.dateType.name === "Operation",
     );
-    const reservationDateRange = season.dateRanges.find(
+    const reservationDateRange = dateRanges.find(
       (dateRange) => dateRange.dateType.name === "Reservation",
     );
```
duncan-oxd, line 136: Should this be a `filter` because there can be more than one reservation date for a dateable?

duncan-oxd (approved):
Looks good! A minor note about a comment to adjust. It would also be good to add a quick comment to any new functions created to say what they do and what they return 👍 Otherwise, looks good to merge!

backend/utils/firstComeFirstServedHelper.js
ayumi-oxd: Good catch, thanks! Updated it.

duncan-oxd (approved):

PR #286, opened 2025-08-13
CMS-904: Update populate previous date script

ayumi-oxd:
### Jira Ticket

CMS-904

### Description

- Additional change on https://github.com/bcgov/bcparks-staff-portal/pull/276
- These dates from the Excel file are past dates, so they were already approved, but haven't been published yet to Strapi.


duncan-oxd (approved):
Makes sense!

PR #289, opened 2025-08-13
CMS-1020: Update date range dropdown

ayumi-oxd:
### Jira Ticket

CMS-1020

### DIFF
--- a/frontend/src/components/DateRangeFields.jsx
+++ b/frontend/src/components/DateRangeFields.jsx
@@ -2,7 +2,7 @@ import { useMemo } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import { faPlus, faXmark } from "@fa-kit/icons/classic/regular";
-import { startOfYear, endOfYear, addYears } from "date-fns";
+import { startOfYear, endOfYear, addYears, addDays } from "date-fns";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import Form from "react-bootstrap/Form";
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/DateRangeFields.jsx
```diff
@@ -20,10 +20,10 @@ function DateRange({
   // A unique ID for template loops and selectors
   const idOrTempId = dateRange.id || dateRange.tempId;
 
-  // Min and max dates: Jan 1 of this year and Dec 31 of next year
+  // Min and max dates: Jan 1 of next year and Dec 31 of the year after next
   // @TODO: Update this when validation is implemented
-  const minDate = useMemo(() => startOfYear(new Date()), []);
-  const maxDate = useMemo(() => endOfYear(addYears(new Date(), 1)), []);
+  const minDate = useMemo(() => startOfYear(addYears(new Date(), 1)), []);
```
duncan-oxd, line 25: This works for a quick fix, but eventually we'll want to use the season `operatingYear` for this, because they might be editing a 2027 season for group camping in 2025 and we'd want this to open up to 2027 👍

duncan-oxd (approved):
I was actually just asking Amanda about the datepickers today! She said there's a ticket in the backlog to set the month that opens when you click it, etc. but this is a nice improvement for now.

In DOOT v1 we'd have the startDate open to January of the operatingYear if there was no value yet, otherwise it just opens to the month of the selected date. The endDate would open to the month of the startDate (or January if there was no value yet).

I'll find the ticket in the backlog and make sure all of the functionality we want is documented in there 👍

PR #290, opened 2025-08-13
CMS-1077: Validate Operating and Reservation dates together

duncan-oxd:
### Jira Ticket

CMS-1077

### Description

This branch builds on #282 to add more validation rules. These new rules check the operating dates against the reservation dates. The rules were mostly the same in v1 of DOOT but I re-implemented them here with the new validation process and the many levels of dateable things.

Note that many of the rules describe the same checks and have the same error message, for example:
> - Operating dates must fully contain reservation dates
> - Reservation dates must be fully contained within operating dates
> - Reservation dates must be within operating dates
> 
> Error message:
> _Enter the reservation dates that fall within the operating dates selected._ 

All three of those rules are describing the same logic, so there's only one check and the message is only shown once.

One other thing I updated in this branch is the "elements" object in the useValidation hook. Some element names are dynamic because they have IDs in them and we were using "magic strings". I added some functions to build the element name strings from their ID inputs. Now every validation rule that calls `addError` and every `ValidationErrorSlot` component will use these functions to get consistent strings for the names.

### DIFF
--- a/frontend/src/components/SeasonForms/AreaSeasonForm.jsx
+++ b/frontend/src/components/SeasonForms/AreaSeasonForm.jsx
@@ -39,7 +39,7 @@ function FeatureFormSectionComponent({
   const { elements } = useValidationContext();
 
   return (
-    <div className="area-feature" key={feature.id}>
+    <div className="area-feature mb-4" key={feature.id}>
       <h4 className="feature-name">{feature.name}</h4>
 
       {featureDateTypes.map((dateType) => (


frontend/src/lib/isDateRangeWithinDateRange.js
duncan-oxd, line 1: date-fns didn't seem to have a function for this. You can check if two ranges overlap, but not if one is "inside" the other.

frontend/src/lib/consolidateDateRanges.js
duncan-oxd, line 1: This cleans up and sorts potentially messy date inputs. This was based on a function we used in v1 and we're using this exact same code on the backend.

ayumi-oxd (approved):
Looks good 👏 I just left a nitpicky comment but it's not a big deal 👍 Code works as expected!

Also, I wasn't sure if it's covered in this ticket or other tickets, but it gives you a "Required" warning when you select the "Date are the same every year" checkbox.

Screenshot 2025-08-14 at 10 26 58 AM

duncan-oxd:
Good catch with the DateRangeAnnual thing - I think that is a problem in CMS-510 so I'll fix it in there (it affects more than just the operating/reservation date types)

PR #291, opened 2025-08-13
CMS-1097: Update filter by date type

ayumi-oxd:
### Jira Ticket

CMS-1097

### Description

- Filter parks by date types, including not only parks/parkAreas/features that have dates with the specific date type, but also those without dates if the following properties are applicable: `park.hasTier1Dates`, `park.hasTier2Dates`, `park.hasWinterFeeDates`, and `feature.hasBackcountryPermits`

### DIFF
--- a/frontend/src/router/pages/EditAndReview.jsx
+++ b/frontend/src/router/pages/EditAndReview.jsx
@@ -206,6 +206,27 @@ function EditAndReview() {
         if (
           filters.dateTypes.length > 0 &&
           !filters.dateTypes.some((filterDateType) => {
+            // check park.hasTier1Dates, park.hasTier2Dates, park.hasWinterFeeDates, and dateTypes
+            if (filterDateType.name === "Tier 1" && !park.hasTier1Dates) {
+              return false;
+            }
+            if (filterDateType.name === "Tier 2" && !park.hasTier2Dates) {
+              return false;
+            }
+            if (
+              filterDateType.name === "Winter fee" &&
+              !park.hasWinterFeeDates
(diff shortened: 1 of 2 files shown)


frontend/src/router/pages/EditAndReview.jsx
```diff
@@ -206,6 +206,27 @@ function EditAndReview() {
...
+            ) {
+              return false;
+            }
+            // check feature.hasBackcountryPermits and dateTypes
+            if (
+              filterDateType.name === "Backcountry registration" &&
+              !park.features?.some((feature) => feature.hasBackcountryPermits)
+            ) {
+              return false;
+            }
```
duncan-oxd, line 228: I guess we don't have info at the Park level for "this park _can_ have backcountry registration dates" so we still have to rely on the Features inside it 👍 But this will work even if those features have no dates, so that's good.

Does this code also check the features within ParkAreas, or just the Features on their own? It looks like we might need to flatten the Park->ParkAreas->Features and the Park->Features and then check that. (The filter is showing 0 results on my local, but I know I have some Features in ParkAreas with `hasBackcountryPermits=true`

duncan-oxd (changes requested):
One small thing to fix, I think!

frontend/src/router/pages/EditAndReview.jsx
ayumi-oxd: Good catch! Added code to check the parkArea > Features date type.

duncan-oxd (approved):
Looks good now!

PR #292, opened 2025-08-14
CMS-970: Create blank Operating dates when Parks have gates

duncan-oxd:
### Jira Ticket

CMS-970

### Description

Added a script to create blank Operating DateRanges for Parks that have `gateDetail.hasGate = true`

I added an association so we can include `GateDetail` in a `Park` query instead of needing a second query, or needing to go through Park->Publishable->GateDetail (now that I know how to do it 😆) 
I didn't change anything for the Dateable at this point though. I'll put something in the backlog to add more associations later to streamline our queries.

I updated the create-blank-date-ranges script to call this new script at the bottom. They do the same thing, but I thought it was good to separate it into its own file because the logic is different. And create-blank-date-ranges gets called in create-seasons.js so basically calling `create-seasons.js` will do everything we need.


ayumi-oxd (approved):
Looks good!

PR #295, opened 2025-08-15
CMS-1020: Use park area publishable id for group camping feature seasons

ayumi-oxd:
### Jira Ticket

CMS-1020

### Description

- When next year's seasons are created (e.g. 2026), the year after the next seasons will be created for the group campsites (e.g. 2027), but these seasons weren't displayed in the table and the form because the season used `feature.publishableId`, not `parkArea.publishableId`
- Updated `create-season` script to create seasons for the group campsites based on `parkArea.publishableId` if its `feature` belongs to `parkArea`
  - If not, create seasons based on `feature.publishableId`
 - Updated the `/park` endpoint to display empty date ranges as "Not provided" for the last year's season in the table
 - For testing, run `node tasks/re-sync-seasons-and-dates/re-sync-season-and-dates.js`

### DIFF
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -84,9 +84,7 @@ function featureModel(minYear, where = {}) {
 // e.g. {Operation: {2024: [...], 2025: [...]}, Winter: {2024: [...], 2025: [...]}, ...}
 function groupDateRangesByTypeAndYear(dateRanges, hasGate = null) {
   // filter out invalid dateRanges
-  let validRanges = dateRanges.filter(
-    (dateRange) => dateRange.dateType && dateRange.startDate,
-  );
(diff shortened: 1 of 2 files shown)


duncan-oxd (approved):
Looks good! Converting those two blocks to functions is a nice improvement in this giant file. Please add some JSDoc comments above the two new functions so they're consistent with the other functions in this file, and then it's good to merge! 

They don't have to be huge, but you can use the comment for `createPublishable` as an example:
```js
/**
 * Creates a new Publishable ID and associates it with the given record, if it doesn't already have one.
 * @param {Park|ParkArea|Feature} record The record to check and update
 * @returns {Promise<number>} The record's Publishable ID
 */
async function createPublishable(record) {
```

PR #299, opened 2025-08-19
CMS-874: Show/hide nav menu items depending on user permissions

molund:
### Jira Ticket

CMS-874

### Description
With this change, users will only see menu nav items corresponding to their assigned keycloak roles.

### DIFF
--- a/frontend-legacy/src/components/composite/responsiveDrawer/responsiveDrawer.js
+++ b/frontend-legacy/src/components/composite/responsiveDrawer/responsiveDrawer.js
@@ -5,6 +5,7 @@ import Toolbar from '@material-ui/core/Toolbar';
 import IconButton from '@material-ui/core/IconButton';
 import Drawer from '@material-ui/core/Drawer';
 import MenuIcon from '@material-ui/icons/Menu';
+import PrivateElement from "../../../auth/PrivateElement";
 import CloseIcon from "@material-ui/icons/Close";
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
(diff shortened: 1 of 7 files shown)


frontend-legacy/src/components/composite/responsiveDrawer/responsiveDrawer.js
```diff
@@ -29,25 +30,32 @@ const ResponsiveDrawer = ({ handleTabChange }) => {
...
-    { value: 3, text: "Dates of Operation", href: "/dates/" },
+  const navItems = [
+    { value: 0, text: "Advisories", allowedRoles: ["submitter", "approver"] },
+    {
+      value: 1,
+      text: "Park Access Status",
+      allowedRoles: ["submitter", "approver"],
+    },
+    { value: 2, text: "Activities & Facilities", allowedRoles: ["approver"] },
+    {
```
ayumi-oxd, line 41: I just forgot why "Park Access Status" is available for both roles, but not "Activities & Facilities". Was it documented somewhere?

ayumi-oxd (approved):
LGTM! Just had one question.

frontend/src/hooks/useAccess.js
```diff
@@ -6,5 +6,9 @@ export default function useAccess() {
   const { roles, checkAccess, logOut, isAuthenticated } =
     useContext(AccessContext);
 
-  return { roles, checkAccess, ROLES, logOut, isAuthenticated };
+  function hasAnyAccess(rolesToCheck) {
+    return rolesToCheck.some((role) => checkAccess(role));
+  }
```
duncan-oxd, line 11: It might make sense to move this into the AccessContext file where `checkAccess` is defined, because they're kind of complementary to each other... but we can make a backlog ticket for refactoring too, given the tight timelines here 👍

frontend-legacy/src/components/composite/responsiveDrawer/responsiveDrawer.js
molund: Park Access Status is view only.  That might have been the reasoning.

PR #301, opened 2025-08-20
CMS-904: Update previous dates json

ayumi-oxd:
### Jira Ticket

CMS-904

### Description

- Updated the previous dates by adding one more day
  - When I converted the Excel file to the JSON file, dates were one day behind from the original file for some reason
 - For updating, run `node tasks/populate-previous-dates/populate-previous-dates.js`


duncan-oxd:
Probably a timezone thing? Until we can fix it to make everything timezone-less, the dates sare stored in the DB as UTC times, but converted to local (pacific) time on the frontend, so that will offset them by 7 or 8 hours

duncan-oxd (approved):

PR #304, opened 2025-08-20
CMS-1117: Update has gate default state

ayumi-oxd:
### Jira Ticket

CMS-1117

### Description

- Update `hasGate` to `allowNull: true`, `defaultValue: null`
- Update AdminJS UI to display "null" if it's not true or false
  - Previously, it displayed "No" (false) even if it was null, and that was confusing
- Update script to import `hasGate` 
   -  Previously, it skipped importing if it was null, and `hasGate` was added as "false" as a default 
 - Update validation to display a warning if `hasGate` is null when a form is saved 
- To update the GateDetails: run `node tasks/create-gate-details/create-gate-details.js`

### DIFF
--- a/backend/tasks/create-gate-details/create-gate-details.js
+++ b/backend/tasks/create-gate-details/create-gate-details.js
@@ -62,11 +62,7 @@ export async function createGateDetailsFromStrapi() {
       const parkOperation = parkOperationByOrcs[park.orcs];
 
       if (parkOperation) {
-        if (parkOperation.hasParkGate === null) {
-          gateDetail.hasGate = null;
-        } else {
-          gateDetail.hasGate = parkOperation.hasParkGate;
-        }
+        gateDetail.hasGate = parkOperation.hasParkGate;
         gateDetail.gateOpenTime =
           parkOperation.gateOpenTime ?? gateDetail.gateOpenTime;
         gateDetail.gateCloseTime =
@@ -125,11 +121,7 @@ export async function createGateDetailsFromStrapi() {
       const subArea = parkOpeationSubAreaById[feature.strapiId];
 
       if (subArea) {
-        if (subArea.hasGate === null) {
-          gateDetail.hasGate = null;
-        } else {
-          gateDetail.hasGate = subArea.hasGate;
-        }
+        gateDetail.hasGate = subArea.hasGate;
         gateDetail.gateOpenTime =
           subArea.gateOpenTime ?? gateDetail.gateOpenTime;
         gateDetail.gateCloseTime =


frontend/src/hooks/useValidation/rules/hasGateChosen.js
```diff
@@ -13,7 +13,7 @@ export default function hasGateChosen(seasonData, context) {
   if (!context.submitted) return;
 
   // hasGate radio buttons are initially null, but Yes or No is required
-  if (current.gateDetail === null) {
+  if (current.gateDetail.hasGate === null) {
```
duncan-oxd, line 16: Thanks for updating this, the radio buttons were initially empty when I wrote this rule function but then they became always true or false after, so this rule wasn't doing anything for a bit!

frontend/src/components/SeasonForms/FeatureSeasonForm.jsx
```diff
@@ -248,7 +248,7 @@ FeatureSeasonForm.propTypes = {
     readyToPublish: PropTypes.bool.isRequired,
     dateRangeAnnuals: PropTypes.arrayOf(PropTypes.object).isRequired,
     gateDetail: PropTypes.shape({
-      hasGate: PropTypes.bool,
+      hasGate: PropTypes.oneOf([true, false, null]),
```
duncan-oxd, line 251: Cool! This is slightly more explicit than saying it's a `bool` but not required. 👍

backend/tasks/create-gate-details/create-gate-details.js
```diff
@@ -62,7 +62,11 @@ export async function createGateDetailsFromStrapi() {
       const parkOperation = parkOperationByOrcs[park.orcs];
 
       if (parkOperation) {
-        gateDetail.hasGate = parkOperation.hasParkGate ?? gateDetail.hasGate;
+        if (parkOperation.hasParkGate === null) {
+          gateDetail.hasGate = null;
+        } else {
+          gateDetail.hasGate = parkOperation.hasParkGate;
+        }
```
duncan-oxd, line 69: Is this the same as just writing `gateDetail.hasGate = parkOperation.hasParkGate;` ?

backend/components/NullableBooleanList.jsx
```diff
@@ -0,0 +1,10 @@
+const NullableBooleanList = ({ record, property }) => {
+  const value = record.params[property.path];
+
+  if (value === null || value === undefined) {
+    return <span>null</span>;
+  }
+  return <span>{value ? "Yes" : "No"}</span>;
```
duncan-oxd, line 7: Cool! This could be handy for any nullable boolean field (if we ever have any more...)

duncan-oxd (approved):
Looks good! I had one thought about optimizing some of the logic in the create-gate-details script, but everything looks good!

backend/tasks/create-gate-details/create-gate-details.js
ayumi-oxd: Good catch! Updated it.

PR #308, opened 2025-08-21
CMS-1185: Add a filter to previous dates at feature in park area level

ayumi-oxd:
### Jira Ticket

CMS-1185

### Description

- Add a `dateableId` filter to previous dates at the feature in the park area level

### DIFF
--- a/frontend/src/components/SeasonForms/AreaSeasonForm.jsx
+++ b/frontend/src/components/SeasonForms/AreaSeasonForm.jsx
@@ -57,7 +57,9 @@ function FeatureFormSectionComponent({
 
           {/* Show previous dates for this featureId/dateableId */}
           <PreviousDates
-            dateRanges={previousFeatureDatesByType?.[dateType.name]}
+            dateRanges={previousFeatureDatesByType?.[dateType.name]?.filter(
+              (dateRange) => dateRange.dateableId === feature.dateableId,
+            )}
           />
 
           <DateRangeFields


frontend/src/components/SeasonForms/AreaSeasonForm.jsx
```diff
@@ -57,7 +57,9 @@ function FeatureFormSectionComponent({
 
           {/* Show previous dates for this featureId/dateableId */}
           <PreviousDates
-            dateRanges={previousFeatureDatesByType?.[dateType.name]}
+            dateRanges={previousFeatureDatesByType?.[dateType.name]?.filter(
+              (dateRange) => dateRange.dateableId === feature.dateableId,
+            )}
```
duncan-oxd, line 62: It might be good to create a memoized map of "previous feature dates by type and dateable" that we can look up in this loop, but this is ok for now 👍 

Do we also need to do this on line 491 where there's another PreviousDates component?

duncan-oxd (approved):
Just one question in case we need to update two different areas. There's some repeated code in this giant file right now 🫠

frontend/src/components/SeasonForms/AreaSeasonForm.jsx
ayumi-oxd: We need this only for the area-feature level dates, because that's the only array that has multiple dates with different dateableId in the group of date type. 
e.g. `{ Operation: [{dateableId: 1, ...}, {dateableId: 2, ...}, ...]}`

I wonder if line 491 is ever going to be rendered since we don't have the area-level dates now.

PR #309, opened 2025-08-21
CMS-1078: Add Tier 1 & 2 validation rules

duncan-oxd:
### Jira Ticket

CMS-1078

### Description

Added 3 more validation rule functions. These ones check if the Reservation dates in a Feature/Area form are exactly equal to the combined Tier 1 and 2 dates for the Park, and vice-versa.

I had to add some metadata to the Park and Area API payloads to include the Park Tier1/2 dates, and include the feature reservation dates in the Park payload.

I still need to confirm some of the logic with Amanda but everything else is ready to review as-is. If I have to change anything, I'll explain it in a comment.

- [x] Confirm reservation dates logic w/ Amanda before merging

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -257,6 +257,64 @@ const SEASON_ATTRIBUTES = [
   "publishableId",
 ];
 
+/**
+ * Returns all reservation feature dates for a specific park and operating year.
+ * @param {Object} park Park model with features and parkAreas
+ * @param {number} operatingYear Operating year for the Seasons
+ * @returns {Promise<Array>} - Array of reservation feature dates
+ */
+async function getFeatureReservationDates(park, operatingYear) {
+  // Don't fetch other dates if the park doesn't have both Tier 1 and Tier 2 dates
+  if (!(park.hasTier1Dates && park.hasTier2Dates)) return [];
+
+  // Get the ID of the applicable Reservation date type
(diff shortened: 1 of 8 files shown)


duncan-oxd:
Confirmed the logic with Amanda and pushed an update!

backend/routes/api/seasons.js
```diff
@@ -257,6 +257,108 @@ const SEASON_ATTRIBUTES = [
...
+  const featurePublishableIds = park.features
+    // Filter out any park features without Publishable IDs
+    .filter((feature) => feature.publishableId)
+    .map((feature) => feature.publishableId);
+
+  const areaFeaturePublishableIds = park.parkAreas
+    // Filter out any park areas without Publishable IDs
+    .filter((parkArea) => parkArea.publishableId)
+    .map((parkArea) => parkArea.publishableId);
+
```
ayumi-oxd, line 288: Do we need to check `park.parkArea.features` too?

ayumi-oxd (approved):
Looks good! Just had one question.

backend/routes/api/seasons.js
duncan-oxd: In this case, we're just using the publishableIds to look up the seasons (for parkAreas and non-area individual Features) - and then we use the season IDs to look up any Reservation date ranges on line 306, which would include all feature dates (inside areas or not) 👍

PR #318, opened 2025-08-28
CMS-1079: Trigger winterDateYears when either date field is filled

duncan-oxd:
### Jira Ticket

CMS-1079

### Description

A small fix for the trigger of the "winter date years" validation rules. 
Previously, it wouldn't check unless both start and end date fields were filled in. As per the requirements, this updates it to check when at least one field is filled in.

I also moved two variable definitions out of the loop so they don't have to get defined over and over again.


ayumi-oxd (approved):
LGTM!

PR #322, opened 2025-09-02
CMS-981: Add export page

ayumi-oxd:
### Jira Ticket

CMS-981

### Description

- Display the export page (tab) for DOOT approver, super admin, and all park access users
- Hide the other filters UI but keep the year filter
- Keep the other filters functionalities and pre-select all values (dateTypes and features)
- Remove pre-selecting a year
- Change HTML select element to react-select <Select> for consistency with other filters on the park page

### DIFF
--- a/backend/routes/api/export.js
+++ b/backend/routes/api/export.js
@@ -75,7 +75,10 @@ router.get(
 
         order: [[Sequelize.col("operatingYear"), "ASC"]],
       })
-    ).map((year) => year.operatingYear);
+    ).map((year) => ({
+      label: year.operatingYear.toString(),
+      value: year.operatingYear,
+    }));
 
     res.json({
       years,
--- a/frontend/src/components/EditAndReviewTable.jsx
+++ b/frontend/src/components/EditAndReviewTable.jsx
@@ -177,8 +177,9 @@ function StatusTableRow({
   // user role
   const { ROLES, checkAccess } = useAccess();
   const approver = useMemo(
-    () => checkAccess(ROLES.APPROVER),
-    [checkAccess, ROLES.APPROVER],
+    () =>
+      checkAccess([ROLES.APPROVER, ROLES.SUPER_ADMIN, ROLES.ALL_PARK_ACCESS]),
+    [checkAccess, ROLES.APPROVER, ROLES.SUPER_ADMIN, ROLES.ALL_PARK_ACCESS],
   );
 
   /**
--- a/frontend/src/components/EditAndReviewTable.jsx
+++ b/frontend/src/components/EditAndReviewTable.jsx
@@ -177,9 +177,8 @@ function StatusTableRow({
   // user role
   const { ROLES, checkAccess } = useAccess();
   const approver = useMemo(
-    () =>
-      checkAccess([ROLES.APPROVER, ROLES.SUPER_ADMIN, ROLES.ALL_PARK_ACCESS]),
-    [checkAccess, ROLES.APPROVER, ROLES.SUPER_ADMIN, ROLES.ALL_PARK_ACCESS],
+    () => checkAccess(ROLES.APPROVER),
+    [checkAccess, ROLES.APPROVER],
   );
 
   /**
--- a/frontend/src/router/layouts/LandingPageTabs.jsx
+++ b/frontend/src/router/layouts/LandingPageTabs.jsx
@@ -34,12 +34,14 @@ export default function LandingPageTabs() {
                   Publish
                 </NavLink>
               </li>
+            )} */}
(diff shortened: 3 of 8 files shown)


backend/routes/api/export.js
```diff
@@ -75,7 +75,10 @@ router.get(
 
         order: [[Sequelize.col("operatingYear"), "ASC"]],
       })
-    ).map((year) => year.operatingYear);
+    ).map((year) => ({
+      label: year.operatingYear.toString(),
+      value: year.operatingYear,
+    }));
```
ayumi-oxd, line 82: I needed to convert an array of numbers to an array of objects for the react-select options prop.

frontend/src/components/EditAndReviewTable.jsx
```diff
@@ -177,8 +177,9 @@ function StatusTableRow({
   // user role
   const { ROLES, checkAccess } = useAccess();
   const approver = useMemo(
-    () => checkAccess(ROLES.APPROVER),
-    [checkAccess, ROLES.APPROVER],
+    () =>
+      checkAccess([ROLES.APPROVER, ROLES.SUPER_ADMIN, ROLES.ALL_PARK_ACCESS]),
+    [checkAccess, ROLES.APPROVER, ROLES.SUPER_ADMIN, ROLES.ALL_PARK_ACCESS],
```
ayumi-oxd, line 182: This "approver" may not be appropriate since it has more than that role. Any suggestions for a name?

frontend/src/styles/_reactselect.scss
```diff
@@ -0,0 +1,27 @@
+.react-select {
+  // Focus state
+  &__control {
+    &--is-focused {
+      border-color: #8099b3 !important;
+      box-shadow: 0 0 0 0.25rem rgba(1, 51, 102, 0.25) !important;
```
ayumi-oxd, line 6: `#8099b3` is used in https://digitalspace.github.io/bcparks-bootstrap-theme/ but not in the BC Gov design token.

frontend/src/router/layouts/LandingPageTabs.jsx
```diff
@@ -11,6 +11,11 @@ export default function LandingPageTabs() {
     () => checkAccess(ROLES.APPROVER),
     [checkAccess, ROLES.APPROVER],
   );
+  // Check if the user has permission to access export page
+  const userExportPermission = useMemo(
+    () => checkAccess([ROLES.APPROVER, ROLES.ALL_PARK_ACCESS]),
+    [checkAccess, ROLES.APPROVER, ROLES.ALL_PARK_ACCESS],
+  );
```
duncan-oxd, line 18: I recently updated `checkAccess` to be a memoized callback, so it won't change unless the roles change (and they will never change during the session). Copilot assured me that we don't need to memoize these checks because it's fairly lightweight (seeing if a value is in a small array) but I can't say that for certain because I'm a React noob 😆 It's all good the way it is, I'm just letting you know what Copilot suggested for future reference.

frontend/src/router/pages/ExportPage.jsx
```diff
@@ -122,7 +134,10 @@ function ExportPage() {
     );
 
   const disableButton =
-    exportFeatures.length === 0 || exportDateTypes.length === 0 || generating;
+    !exportYear ||
+    exportFeatures.length === 0 ||
+    exportDateTypes.length === 0 ||
```
duncan-oxd, line 139: Let's get rid of these two checks while the form elements are hidden. We might not end up using the featureType option on this form now that we have Park-level dates, etc.

frontend/src/router/pages/ExportPage.jsx
```diff
@@ -16,54 +17,65 @@ function ExportPage() {
...
+  const exportType = "all";
 
   const { generating, fetchData: fetchCsv } = useApiGet("/export/csv", {
     instant: false,
     params: {
       type: exportType,
-      year: exportYear,
+      year: exportYear?.value,
       "features[]": exportFeatures,
       "dateTypes[]": exportDateTypes,
```
duncan-oxd, line 36: Let's just send the `year` value here. I have a suspicion this page will change, but it won't go back to the way it looked before, so it should be okay to totally delete the stuff we're hiding.

duncan-oxd (changes requested):
Looks good! Just a small note about border-radius styles, and some thoughts about removing stuff instead of just hiding it. No need to update the API endpoint if it doesn't work in this branch, because I've done that in my PR. Just trying to avoid creating more technical debt in the form of comments to remove some day 😄

duncan-oxd (approved):

PR #324, opened 2025-09-04
CMS-1015: Disable approve button when not applicable

ayumi-oxd:
### Jira Ticket

CMS-1015

### Description

- Disable the approve button in the park table when the status is Approved/Published/Requested by HQ

### DIFF
--- a/frontend/src/components/EditAndReviewTable.jsx
+++ b/frontend/src/components/EditAndReviewTable.jsx
@@ -136,9 +136,7 @@ DateTableRow.propTypes = {
 
 function ApproveButton({ seasonId, status, color = "", onApprove }) {
   // disable the approve button if a season is already approved, published, or requested by HQ
-  const isDisabled = ["approved", "published", "requested"].includes(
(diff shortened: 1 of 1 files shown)


frontend/src/components/EditAndReviewTable.jsx
```diff
@@ -125,7 +134,9 @@ DateTableRow.propTypes = {
   currentYear: PropTypes.number,
 };
 
-function ApproveButton({ seasonId, color = "", onApprove }) {
+function ApproveButton({ seasonId, status, color = "", onApprove }) {
+  // disable the approve button if a season is already approved, published, or requested by HQ
+  const isDisabled = status !== "submitted";
```
duncan-oxd, line 139: With this simpler logic, we _could_ change it so `disabled` is the prop, and we just pass in the value of `status !== "submitted"` but then it would be more work to update if if the logic ever has to change. So it's probably best to leave it the way it is, but I'm just thinking out loud 😅

duncan-oxd (approved):

PR #328, opened 2025-09-10
CMS-1194: Filter nextYearFeatures for in inReservationSystem

duncan-oxd:
### Jira Ticket

CMS-1194

### Description

An extra filter for the features that get seasons requested for "the next year" (ie 2027) - it should only be for features in the BCP reservation system.

This is a 2.0.1 ticket, so I'll move it down in the git history so it can come right after the current v2.0.1-uat.1 tag. It shouldn't cause any conflicts because it just changes the one line in one backend file, but I'll give everyone a heads up before I do that 😛

### DIFF
--- a/backend/tasks/create-seasons.js
+++ b/backend/tasks/create-seasons.js
@@ -392,8 +392,10 @@ seasonsAdded = 0;
 
 // Find all features that need to request dates for next year
 const nextYearFeatures = await Feature.findAll({
+  // Only active features in the BC Parks reservation system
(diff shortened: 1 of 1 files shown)


ayumi-oxd (approved):
[This PR](https://github.com/bcgov/bcparks-staff-portal/pull/329) is technically v.2.1.0, but it'd be great if you could include it with the v.2.0.1-uat along with your PR, given it has been passed to UAT already 🙏

PR #337, opened 2025-09-15
CMS-1229: Update admin js relations

ayumi-oxd:
### Jira Ticket

CMS-1229

### Description
- Add ths adminjs/relations plugin
- Restructure table indexes to no longer use natural keys

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -46,6 +46,7 @@
     "express-jwt": "^8.4.1",
     "express-rate-limit": "^7.4.0",
     "express-session": "^1.18.0",
+    "flat": "^5.0.2",
     "helmet": "^7.1.0",
     "jwks-rsa": "^3.1.0",
     "lodash": "^4.17.21",
--- a/backend/components/JsonEdit.jsx
+++ b/backend/components/JsonEdit.jsx
@@ -29,11 +29,18 @@ function JsonEdit(props) {
     setValue(val);
 
     try {
-      const parsed = val.trim() ? JSON.parse(val) : null;
+      if (val.trim() === "") {
+        // Send null for empty values
+        onChange(property.path, null);
+        setError("");
+      } else {
+        // Parse to validate, but send as JSON string to preserve types
+        JSON.parse(val);
 
-      // Notify AdminJS of the change
-      onChange(property.path, parsed);
-      setError("");
+        // Send the raw JSON string with a special marker
+        onChange(property.path, `__JSON_STRING__${val}`);
+        setError("");
+      }
     } catch (err) {
       // keep showing error until valid JSON
       console.error("JSON parse error:", err);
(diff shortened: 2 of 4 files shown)


backend/.env.local.example
```diff
@@ -17,3 +17,6 @@ ADMIN_SESSION_SECRET="another random string"
 STRAPI_TOKEN="your-strapi-token"
 # URL of the Strapi API
 STRAPI_URL="http://host.docker.internal:1337"
+
+# Find license key in the 1password vault
+ADMINJS_RELATIONS_LICENSE_KEY="adminjs-relations-license-key"
```
molund, line 22: This variable should be added (as blank) to the helm config for the backend.

backend/migrations/20250912161835-update-non-pk-foreign-keys.js
```diff
@@ -0,0 +1,141 @@
+/** @type {import('sequelize-cli').Migration} */
+/** @type {import('sequelize-cli').Migration} */
```
duncan-oxd, line 2:
```suggestion
/** @type {import('sequelize-cli').Migration} */
```
Duplicated line

backend/index.js
```diff
@@ -31,6 +31,9 @@ if (process.env.NODE_ENV === "production") {
   app.set("trust proxy", 1);
 }
 
+// AdminJS routes
+app.use(admin.options.rootPath, adminRouter);
```
duncan-oxd, line 35: Discussed on slack about trying to fix up the middleware so we don't have to move the route definition up

backend/package.json
```diff
@@ -17,6 +17,9 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@babel/core": "^7.28.4",
+    "@babel/eslint-parser": "^7.28.4",
+    "@babel/preset-react": "^7.27.1",
```
duncan-oxd, line 22: I don't know why these dev dependencies would be required for the backend, but they're harmless even if we're not using them...

duncan-oxd (commented):
A few comments as we're all going through this on Slack. One's a suggestion that I'll go ahead and apply myself

backend/index.js
```diff
@@ -32,7 +33,6 @@ if (process.env.NODE_ENV === "production") {
 }
 
 // JSON parsing middleware
-app.use(express.json());
```
duncan-oxd, line 36:
```suggestion
```
If this change is permanent, we can move the comment down too

backend/index.js
```diff
@@ -56,6 +56,12 @@ const limiter = RateLimit({
 
 app.use(limiter);
 
+// AdminJS routes
+app.use(admin.options.rootPath, adminRouter);
+
+// app.use(express.json());
```
duncan-oxd, line 62:
```suggestion
// JSON parsing middleware
```

backend/.env.local.example
duncan-oxd: I added it to the adminjs-secret and updated the docs 👍

PR #342, opened 2025-09-19
CMS-1238: Update gate detail logic

ayumi-oxd:
### Jira Ticket

CMS-1238

### Description

- Filter out rows where `hasGate` is false and `dateType` is "Operating (Gate)"
- Not export gate times or annual status if `hasGate` is false

### DIFF
--- a/backend/routes/api/export.js
+++ b/backend/routes/api/export.js
@@ -462,7 +462,7 @@ router.get(
           gateStartTime,
           gateEndTime,
           sameEveryYear: gateSameEveryYear,
-        } = getGateDisplayValues(dateRange, season.gateDetail, annualData);
+        } = getGateDisplayValues(dateRange, gateDetail, annualData);
 
         // Get the final sameEveryYear value
         // For gate (operating) date type, use the value from getGateDisplayValues
@@ -495,7 +495,7 @@ router.get(
           [colNames.START_DATE]: formatDate(dateRange.startDate),
           [colNames.END_DATE]: formatDate(dateRange.endDate),
           [colNames.SAME_EVERY_YEAR]: sameEveryYear,
-          [colNames.HAS_GATE]: formatBoolean(gateDetail?.hasGate),
+          [colNames.HAS_GATE]: formatBoolean(hasGate),
           [colNames.GATE_START_TIME]: gateStartTime,
           [colNames.GATE_END_TIME]: gateEndTime,
           [colNames.IN_BCP_RESERVATION_SYSTEM]: formatBoolean(
--- a/backend/routes/api/export.js
+++ b/backend/routes/api/export.js
@@ -238,12 +238,13 @@ function getFeatureForDateRange(dateRange) {
 
 /**
  * Calculates the GateDetail display values for a DateRange.
+ * @param {DateRange} dateRange The DateRange object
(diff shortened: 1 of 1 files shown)


backend/routes/api/export.js
```diff
@@ -416,6 +416,32 @@ router.get(
...
+          gateStartTime = gateDetail?.gateOpensAtDawn
+            ? "Opens at dawn"
+            : formatTime(gateDetail?.gateOpenTime);
+
+          gateEndTime = gateDetail?.gateClosesAtDusk
+            ? "Closes at dusk"
+            : formatTime(gateDetail?.gateCloseTime);
+
+          // Only show isDateRangeAnnual if hasGate is true
+          sameEveryYear = formatBoolean(annualData?.isDateRangeAnnual);
```
duncan-oxd, line 437: If this isn't a gate-type date, do we still want to show true or false for this? For example, the operating/reservation dates for features/campgrounds in the screenshot below:

Image

backend/routes/api/export.js
```diff
@@ -416,6 +416,32 @@ router.get(
         // Get the Feature for this DateRange, if there is one
         const feature = getFeatureForDateRange(dateRange);
 
+        // Pre-calculate GateDetail
+        const isGateType = dateRange.dateType.name === "Operating (Gate)";
+        const hasGate = gateDetail?.hasGate === true;
+
+        let gateStartTime = "";
+        let gateEndTime = "";
+        let sameEveryYear = "";
```
duncan-oxd, line 425: How about moving this logic to a dedicated function, similar to wh at we've done with `getFeatureForDateRange` and `getInReservationSystem`?

It could return an object with all of the values we need for this:

```suggestion
        // something like this
        const { gateStartTime, gateEndTime, sameEveryYear } = getGateDetails(season);
```

duncan-oxd (changes requested):
Thanks for taking care of this! I added a question and a small change request 👍

backend/routes/api/export.js
ayumi-oxd: Good catch! I believe this calculation should be for the `dateType` "Operating". For other date types, it should display as is. I added a condition in the function.

duncan-oxd (changes requested):
A few changes to clean it up a little: move some more of the gate-related stuff into the function, and possibly move the dateRangeAnnual stuff out of the function, because it's not completely gate-related. Let me know if you have questions or concerns!

backend/routes/api/export.js
```diff
@@ -416,6 +455,26 @@ router.get(
...
         const feature = getFeatureForDateRange(dateRange);
 
+        // Get the GateDetail display values
+        const {
+          isGateType,
+          hasGate,
+          gateStartTime,
+          gateEndTime,
+          sameEveryYear: gateSameEveryYear,
+        } = getGateDisplayValues(dateRange, season.gateDetail, annualData);
```
duncan-oxd, line 465:
```suggestion
        } = getGateDisplayValues(dateRange, gateDetail, annualData);
```
We already have `gateDetail` from the section above, so we could simplify this, but otherwise this looks perfect 👍

duncan-oxd (approved):
Nice! I put a small note if you want to make one more tweak, but it's ready to merge 👍

PR #346, opened 2025-09-23
CMS-000: Hide park area if it does not have active features

ayumi-oxd:
### Jira Ticket

NOBUG
### Description

- Hide park area if it does not have active features at `/park`, since the park area itself won't have dates


duncan-oxd (approved):
Looks good! I'm gonna change the title to "CMS-000:" just for consistency with other commits in this repo that didn't correspond to a jira ticket

PR #347, opened 2025-09-26
CMS-899: Add publish page

ayumi-oxd:
### Jira Ticket

CMS-899

### Description

- Add the publish page back for HQ stuff
- Update the API `/ready-to-publish` endpoint using the v2 data model
- Update the publish table and display park, area, features, and year
- Update populatePreviousDates to change 2025 Tier 1& 2 dates season status "approved" to "published" so that there will be no seasons on the publish page
  - For testing, run `node tasks/populate-previous-dates/populate-previous-dates.js`

### DIFF
--- a/backend/routes/api/publish.js
+++ b/backend/routes/api/publish.js
@@ -107,9 +107,19 @@ router.get(
       }),
       ParkArea.findAll({
         where: { publishableId: { [Op.in]: publishableIds } },
+        attributes: ["id", "publishableId", "name"],
+        include: [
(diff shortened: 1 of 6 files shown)


duncan-oxd (changes requested):
Nice! I had a few ideas for readability/maintainability tweaks, but it's looking great.

We can probably get all of the "ready to publish" stuff in one query with subqueries, but I think doing 3 queries with Promise.all will be fine here, especially if our goal is to get a flattened structure with Publishable IDs in the end 👍

PR #350, opened 2025-09-29
CMS-1265: Delete reservation dates if hasReservations is false

ayumi-oxd:
### Jira Ticket

CMS-1265

### Description

- Some park-operation-sub-areas `hasReservations` are updated from true to false recently, but DOOT still has "Reservation" dates in these features (sub-areas), so we need to sync `hasReservations` again and delete "Reservation" dates if it's false
- For testing:
  - Run `node strapi-sync/import-data.js` to update Feature `hasReservations`
  - Run `node tasks/delete-reservation-date-ranges/delete-reservation-date-ranges.js` to delete existing "Reservation" dates if Feature's `hasReservations` is false 
 - @duncan-oxd Do you happen to remember how "Reservation" dates are created? Is it based on `hasReservations`  and the past dates? It might be related to CMS-970, so I wanted to ask you if you know 🙂

### DIFF
--- a/backend/tasks/delete-reservation-date-ranges/delete-reservation-date-ranges.js
+++ b/backend/tasks/delete-reservation-date-ranges/delete-reservation-date-ranges.js
@@ -1,0 +1,63 @@
+// This script deletes reservation DateRanges for parks/features where hasReservations is false
+
+import "../../env.js";
+
+import { Op } from "sequelize";
+import { DateType, DateRange, Feature } from "../../models/index.js";
+
+async function deleteReservationDateRanges(transaction = null) {
+  try {
+    // Find the reservation DateType id
+    const reservationDateType = await DateType.findOne({
+      where: { name: "Reservation" },
+      attributes: ["id"],
+      transaction,
+    });
+
+    if (!reservationDateType) {
+      throw new Error("No Reservation DateType found");
+    }
+
+    // Find all features with hasReservations false
+    const features = await Feature.findAll({
+      where: { hasReservations: false },
+      attributes: ["dateableId"],
+      transaction,
+    });
+    const featureDateableIds = features.map((feature) => feature.dateableId);
+
+    // Delete reservation DateRanges for features
+    const deleteCount = await DateRange.destroy({
+      where: {
+        dateTypeId: reservationDateType.id,
+        dateableId: {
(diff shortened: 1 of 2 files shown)


duncan-oxd:
> Do you happen to remember how "Reservation" dates are created? Is it based on `hasReservations`  and the past dates?

That's correct, the script that creates the blank dates (ie, "requests dates for a new season") uses the `getDateTypesForFeature` function to get the applicable date types for a feature, and all it does for reservations is checks the `hasReservations` attribute 👍

ayumi-oxd:
> This looks good, but just note that it will delete all reservation dates for the Feature across _all seasons_. I assume that's what we want, but I just wanted to confirm. If a feature had reservations for a few years and then changed, would we want to keep the old data in the DOOT db?

We've decided not to delete all reservation dates across all seasons. The script only deletes the Reservation dates from the latest/most recent season.

backend/tasks/delete-reservation-date-ranges/delete-reservation-date-ranges.js
```diff
@@ -31,19 +31,65 @@ async function deleteReservationDateRanges(transaction = null) {
...
+    // Group by feature (dateableId), find max operatingYear per feature
+    const latestDateRangeIds = [];
+    const grouped = {};
+
+    dateRanges.forEach((dateRange) => {
+      const key = dateRange.dateableId;
+
+      if (!grouped[key]) grouped[key] = [];
+      grouped[key].push(dateRange);
+    });
```
duncan-oxd, line 61: Don't forget you can use lodash groupBy to do this with one line 👍 
```js
const grouped = groupBy(dateRanges, "dateableId");
```

duncan-oxd (changes requested):
The changes look good! Just a couple of small readability suggestions, and one question I mentioned on Slack:

> Just wanna confirm that we want to delete data for the latest season for all of the date ranges, and not just any that match a specific season (like "2026")
I don't know if it would make any difference at this point, but I'm just thinking if there was data for 2024, then they didn't collect dates for 2025, would it delete 2024?

duncan-oxd (changes requested):
Looks good to me! Let's just update the readme to reflect the logic change and then it's good to merge

PR #476, opened 2026-03-19
CMS-1609: Fix the publishing error if no gate exists

ayumi-oxd:
### Jira Ticket

CMS-1609

### Description

- Issue: Unable to publish tier and gate data if no gate exists
- Changed default parameter `gateDetails = {}` to const `gate = gateDetails ?? {}` inside the function since JavaScript default parameters only fire for undefined, not null.
- Added `getApplicableParkDateTypeIds`: Determines which date type IDs actually apply to a park before fetching anything to prevent demanding dates that the park isn't configured to have.
- Updated `formatDateRanges`: Added `allowedDateTypeIds` filter and `allowEmpty` flag so parks with no applicable date types can return `[]` instead of throwing "No date ranges found".

### DIFF
--- a/backend/routes/api/publish.js
+++ b/backend/routes/api/publish.js
@@ -381,14 +381,23 @@ function getApplicableParkDateTypeIds(park, season) {
  * @throws {Error} If no valid date ranges are found
  */
 async function formatDateRanges(entity, season, options = {}) {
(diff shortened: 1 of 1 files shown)


duncan-oxd (changes requested):
Nice 👍 Just an errant backslash to remove on one line, I think!

PR #499, opened 2026-04-06
CMS-1653: Make activities and facilities page responsive

ayumi-oxd:
### Jira Ticket

CMS-1653

### Description

- Previously the park info page from the activities and facilities page didn't have the side menu, but the new version has always has the menu so the break point need to be changed `lg` to `xl`

### DIFF
--- a/frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
+++ b/frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
@@ -649,28 +649,28 @@ export default function ParkInfo() {
                       {parkActivities && parkActivities.length > 0 && (
                         <div>
                           <div className="row pt2b2 mx-0">
-                            <div className="col-lg-3 col-md-12 col-12 park-header">
+                            <div className="col-xl-3 col-12 park-header">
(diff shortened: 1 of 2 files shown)


frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
```diff
@@ -649,28 +649,28 @@ export default function ParkInfo() {
...
-                            <div className="col-lg-1 col-md-12 col-12 park-header">
+                            <div className="col-xl-1 col-12 park-header">
                               Display
                             </div>
-                            <div className="col-lg-1 col-md-12 col-12 park-header">
+                            <div className="col-xl-1 col-12 park-header">
                               Open
                             </div>
-                            <div className="col-lg-3 col-md-12 col-12 park-header">
+                            <div className="col-xl-3 col-12 park-header">
```
molund, line 661: We should ask Amanda if we can delete this column from all three tabs. It doesn't appear to do anything and our data model doesn't even associate fees with activities, facilities or camping types.

molund (approved):

PR #500, opened 2026-04-06
CMS-1659: Memoized react select options

ayumi-oxd:
### Jira Ticket

CMS-1659

### Description

- Fixed the dropdown options were being rebuilt on every render
- Used `useMemo` for the react select options to avoid re-rendering


duncan-oxd (approved):
Looks good. Note that the dependencies `regions` and `protectedAreas` are re-fetched inside a `useEffect` that triggers whenever the keycloak token changes (every 5 minutes). We can fix this by using the new `useCms` hook instead of directly passing the keycloak token to Axios

PR #502, opened 2026-04-07
NOBUG: Change port to 3000 and add not-so-secret values to env.example

molund:
### Jira Ticket

None

### Description
- Changed the port for the staff portal on localdev from 8101 to 3000. 
    - 3000 was the port used by the legacy staff portal and it is the port used by Caddy on production. 
    - 8101 was the DOOT port
- Added some values to .env.example files that aren't really secrets to simplify dev environment setup
- Add aliases so we can use `npm run develop` instead of `npm run dev` without having to think about which project we are working on (the CMS project uses `develop`)


ayumi-oxd (commented):
LGTM!

ayumi-oxd (approved):

PR #505, opened 2026-04-07
CMS-1635: Use activityName in Park Info table

duncan-oxd:
### Jira Ticket

CMS-1635

### Description

Updates the "Facilities & Activites" table to use the `activityType.activityName` value instead of the `name`, so it won't show the orcs number in that string. As long as `activityType.activityName` is always defined (and correct) this should work as a fix.

### DIFF
--- a/frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
+++ b/frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
@@ -666,9 +666,12 @@ export default function ParkInfo() {
                             </div>
                           </div>
                           {parkActivities.map((a) => (
-                            <div className="row pt2b2 mx-0" key={`activity-${a.id}`}>
+                            <div
+                              className="row pt2b2 mx-0"
+                              key={`activity-${a.id}`}
+                            >
                               <div className="col-xl-3 col-12 park-content">
-                                {a.name}
+                                {a.activityType.activityName}
                               </div>
                               <div className="col-xl-1 col-12 park-content">
                                 <SwitchButton
@@ -821,9 +824,12 @@ export default function ParkInfo() {
                             </div>
                           </div>
                           {parkFacilities.map((f) => (
-                            <div className="row pt2b2 mx-0" key={`facility-${f.id}`}>
+                            <div
+                              className="row pt2b2 mx-0"
+                              key={`facility-${f.id}`}
+                            >
                               <div className="col-xl-3 col-12 park-content">
-                                {f.name}
+                                {f.facilityType.facilityName}
                               </div>
                               <div className="col-xl-1 col-12 park-content">
                                 <SwitchButton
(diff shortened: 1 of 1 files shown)


frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
```diff
@@ -666,9 +666,12 @@ export default function ParkInfo() {
...
                           </div>
                           {parkActivities.map((a) => (
-                            <div className="row pt2b2 mx-0" key={`activity-${a.id}`}>
+                            <div
+                              className="row pt2b2 mx-0"
+                              key={`activity-${a.id}`}
+                            >
                               <div className="col-xl-3 col-12 park-content">
-                                {a.name}
+                                {a.activityType.activityName}
```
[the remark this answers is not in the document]
duncan-oxd: I wondered about that, too. The query makes it look like it will always be populated. Can anyone confirm?

frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
```diff
@@ -821,7 +824,10 @@ export default function ParkInfo() {
                             </div>
                           </div>
                           {parkFacilities.map((f) => (
-                            <div className="row pt2b2 mx-0" key={`facility-${f.id}`}>
+                            <div
+                              className="row pt2b2 mx-0"
+                              key={`facility-${f.id}`}
+                            >
                               <div className="col-xl-3 col-12 park-content">
                                 {f.name}
```
ayumi-oxd: Could you apply the change to facilities too?

frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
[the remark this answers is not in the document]
ayumi-oxd: activityType is required relation in parkActivity, so it always has a type.

ayumi-oxd (commented):
Just left some comments, but looks good overall 👍

frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
duncan-oxd: 😅 I just noticed the Facilities tab. I'll update that.

frontend/src/router/pages/advisories/parkInfo/ParkInfo.jsx
molund: This should be updates as well so Camping Types, Activities and Facilities all follow the same pattern.

duncan-oxd:
Pushed an update so all three tables get their name column from the display name field in the CMS, and not the one with the orcs and the colon. Camping type was already splitting it out, but I switched all three.

ayumi-oxd (approved):
Looks good!

molund (approved):

PR #509, opened 2026-04-08
CMS-1676: Change select option z-index

ayumi-oxd:
### Jira Ticket
CMS-1676
CMS-1660

### Description

- CMS-1676: Changed select option z-index to avoid layer issues
- CMS-1660: Aligned checkbox with show archived label


duncan-oxd (approved):
Changing it _from_ 9999 to 999? 
Looks good to me, but should we add a global style to affect all Select box components?

ayumi-oxd:
> Changing it _from_ 9999 to 999? Looks good to me, but should we add a global style to affect all Select box components?

Unfortunately I haven't found an easy way to apply a class name to this dropdown elements. Any feedback would be appreciated 🙏 

Screenshot 2026-04-08 at 5 12 14 PM

PR #515, opened 2026-04-09
CMS-1678: Fix toast message position

ayumi-oxd:
### Jira Ticket

CMS-1678

### Description

- Fixed toast message position


duncan-oxd (approved):
I didn't know we were using this Toast component anywhere. We can probably use that instead of our homemade flash message thing some day 👍

PR #517, opened 2026-04-13
CMS-1616: Add server side pagination to advisory table

ayumi-oxd:
### Jira Ticket

CMS-1616

### Description

- Added server side pagination, filtering, and sorting to the advisory table (not park access status table)
- Created `AdvisoryDashboardQuery.js` for filters and sorts
- Please note that the region filter works only if advisories has a region
  - It doesn't get advisories with protectedAreas in a region
  - Note: I tried to get them using `$in` https://docs.strapi.io/cms/api/rest/filters#example-find-multiple-restaurants-with-ids-3-68 but it didn't work probably because the API post can be insanely long if a region has more than 100 protectedAreas ...

### DIFF
--- a/frontend/src/components/advisories/composite/dataTable/DataTable.jsx
+++ b/frontend/src/components/advisories/composite/dataTable/DataTable.jsx
@@ -300,6 +300,18 @@ export default function DataTable(props) {
 
   const totalItems = options.totalItems ?? sortedRows.length;
 
+  // In server-side mode, the parent already sends only the current page's rows.
+  // In client-side mode, slice sortedRows to the current page/pageSize here.
+  const displayedRows = useMemo(() => {
+    if (options.serverSide || currentPageSize < 0) {
+      return sortedRows;
+    }
+
+    const start = (currentPage - 1) * currentPageSize;
+
+    return sortedRows.slice(start, start + currentPageSize);
+  }, [currentPage, currentPageSize, options.serverSide, sortedRows]);
+
   function handleSort(column, index) {
     if ((!column.field && !column.customSort) || column.sorting === false) {
       return;
@@ -482,7 +494,7 @@ export default function DataTable(props) {
                 </td>
               </tr>
             )}
-            {sortedRows.map((row, rowIndex) => {
+            {displayedRows.map((row, rowIndex) => {
               const rowKey = row.documentId || row.id || rowIndex;
               const rowProps = {};
 
--- a/frontend/src/components/advisories/composite/dataTable/DataTable.jsx
+++ b/frontend/src/components/advisories/composite/dataTable/DataTable.jsx
@@ -194,7 +194,18 @@ export default function DataTable(props) {
   const {
     columns,
     data,
-    options,
+    exportMenu,
+    filtering,
+    pageSize: pageSizeProp,
+    pageSizeOptions: pageSizeOptionsProp,
+    search,
+    serverSide,
+    totalItems: totalItemsProp,
+    currentPage: currentPageProp,
+    onPageChange,
+    onPageSizeChange: onPageSizeChangeProp,
+    onSortChange,
+    onFilterChange,
     title,
     onRowClick,
     initialFilterValues,
@@ -211,13 +222,13 @@ export default function DataTable(props) {
   const [filterValues, setFilterValues] = useState(initialFilterValues || {});
   const [sortConfig, setSortConfig] = useState(null);
   const [page, setPage] = useState(1);
-  const initialPageSize = options.pageSize || 5;
+  const initialPageSize = pageSizeProp || 5;
   const [pageSize, setPageSize] = useState(initialPageSize);
-  const hasSort = Boolean(options.onSortChange);
-  const currentPage = options.currentPage ?? page;
-  const currentPageSize = options.pageSize ?? pageSize;
+  const hasSort = Boolean(onSortChange);
+  const currentPage = currentPageProp ?? page;
+  const currentPageSize = pageSizeProp ?? pageSize;
   const pageSizeOptions = useMemo(() => {
-    const configured = options.pageSizeOptions || [5, 10, 25, 50, 100];
+    const configured = pageSizeOptionsProp || [5, 10, 25, 50, 100];
     const combined = [...configured, initialPageSize].filter(
       (value) => value !== 0,
     );
@@ -231,7 +242,7 @@ export default function DataTable(props) {
       if (right < 0) return -1;
       return left - right;
     });
-  }, [initialPageSize, options.pageSizeOptions]);
+  }, [initialPageSize, pageSizeOptionsProp]);
 
   useEffect(() => {
     setPageSize(initialPageSize);
@@ -239,12 +250,12 @@ export default function DataTable(props) {
 
   const filteredRows = useMemo(() => {
     // Skip local filtering when the parent handles it server-side
-    if (options.serverSide) {
+    if (serverSide) {
       return data;
     }
 
     return data.filter((row) => {
-      const matchesSearch = options.search
+      const matchesSearch = search
         ? rowMatchesSearch(row, visibleColumns, searchText)
         : true;
       const matchesFilters = visibleColumns.every((column, index) => {
@@ -258,14 +269,14 @@ export default function DataTable(props) {
   }, [
     data,
     filterValues,
-    options.search,
-    options.serverSide,
+    search,
+    serverSide,
     searchText,
     visibleColumns,
   ]);
 
   const sortedRows = useMemo(() => {
-    if (!sortConfig || options.serverSide) {
+    if (!sortConfig || serverSide) {
       return filteredRows;
     }
 
@@ -290,7 +301,7 @@ export default function DataTable(props) {
     });
 
     return nextRows;
-  }, [filteredRows, options.serverSide, sortConfig, visibleColumns]);
+  }, [filteredRows, serverSide, sortConfig, visibleColumns]);
 
   useEffect(() => {
     const totalPages = Math.max(1, Math.ceil(sortedRows.length / pageSize));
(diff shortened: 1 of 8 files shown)


frontend/src/lib/advisories/utils/AdvisoryDataUtil.js
```diff
@@ -17,9 +17,11 @@ export function updatePublicAdvisories(publicAdvisories, managementAreas) {
 
   return publicAdvisories.map((publicAdvisory) => {
     publicAdvisory.expired = publicAdvisory.expiryDate < today ? "Y" : "N";
+    // Display associated parks/regions, or rec resource name if no parks/regions
     publicAdvisory.associatedParks =
       publicAdvisory.protectedAreas.map((p) => p.protectedAreaName).join(", ") +
-      publicAdvisory.regions.map((r) => r.regionName).join(", ");
+        publicAdvisory.regions.map((r) => r.regionName).join(", ") ||
+      publicAdvisory.recreationResources.map((r) => r.resourceName).join(", ");
```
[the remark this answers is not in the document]
ayumi-oxd: `regions` are meant for parks, not for resources. Resources will have `districts` in the future tickets.

frontend/src/components/advisories/composite/dataTable/DataTable.jsx
```diff
@@ -213,38 +213,57 @@ export default function DataTable(props) {
...
     );
+    const unique = [...new Set(combined)];
 
-    return [...new Set(combined)].sort((left, right) => left - right);
+    return unique.sort((left, right) => {
+      // "All" (-1) sorts to end
+      if (left < 0) return 1;
+      if (right < 0) return -1;
+      return left - right;
+    });
```
duncan-oxd, line 233: Can you add some comments to explain what this is doing?

frontend/src/router/pages/advisories/advisory/Advisory.jsx
duncan-oxd, line 540: This is already gone in #514 so I think this branch just needs to be rebased 👍

frontend/src/hooks/useCms.js
```diff
@@ -68,6 +68,19 @@ export default function useCms() {
     [withAuthorization],
   );
 
+  /**
+   * Get full response payload including meta.pagination.
+   * Use this instead of cmsGet when you need pagination metadata.
+   */
+  const cmsGetRaw = useCallback(
```
duncan-oxd: This is a good idea, but we could even tweak the existing `cmsGet` to allow this with a parameter. How about using [lodash.get](https://lodash.com/docs/4.18.1#get) to let us customize the `res.data.data` path it returns?

```js
import { get as lodashGet } from "lodash-es";

// Something like this:
// so it will default to returning res.data.data but you can also customize it for anything

  /**
   * Performs a GET request to the CMS and returns the nested data at the provided `path`.
   * @param {string} url CMS endpoint to fetch
   * @param {Object} config axios.get request config
   * @param {string|Array} [path="data"] lodash.get path to the nested data in the response
   * @returns {Promise<any>} The nested data at the provided path in the CMS response
   */
  const cmsGet = useCallback(
    async (url, config = {}, path = "data") => {
      // Get the JSON data from the response
      const { data: responseData } = await cmsAxios.get(
        url,
        withAuthorization(config),
      );

      // If no path is provided ("" or empty array), return the full response data
      if (path.length === 0) {
        return responseData;
      }

      // Return the nested data at the provided path
      return lodashGet(responseData, path);
    },
    [withAuthorization],
  );
```

So in your case you could get the full response like this:

```js
const countResult = await cmsGetRaw(
  `/public-advisory-audits?${countQuery}`,
  {}, // second param is the axios config, this is usually empty
  "" // empty lodash.get path to just return the axios res.data instead of res.data.data
);

File: frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx [hunk 4]
```diff
@@ -40,6 +44,9 @@ import { updatePublicAdvisories } from "@/lib/advisories/utils/AdvisoryDataUtil"
  * @param {any} [defaultValue=0] Default value to return if the filter isn't found in storage
  * @returns {any} The filter value, or default if not found
  */
+const ALL_PAGE_SIZE = -1;
+const DEFAULT_PAGE_SIZE = 50;
+
```
duncan-oxd: These variable definition lines should go above the jsdoc comment for `getPageFilterValue` - it looks like they got added in between

duncan-oxd (changes requested):
Cool! I requested a few small updates but nothing major 👍

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -337,94 +157,62 @@ export default function AdvisoryDashboard() {
...
+        setManagementAreas(managementAreasData);
+        setProtectedAreas(protectedAreasData);
+
+        // Fetch advisory statuses and urgencies for filter options and table icons
         setAdvisoryStatuses(fetchedAdvisoryStatuses);
         setUrgencies(fetchedUrgencies);
 
-        if (fetchedAdvisoryStatuses) {
+        // Fetch the list of advisory numbers that currently have a live PUB version
+        if (fetchedAdvisoryStatuses?.length > 0) {
```
duncan-oxd: This is after an `await` so we know that `fetchedAdvisoryStatuses` will always be defined as an array at this point and we don't necessarily need the `?`

frontend/src/hooks/useCms.js
```diff
@@ -58,12 +59,25 @@ export default function useCms() {
 
   // Authenticated wrappers around cmsAxios methods
 
+  /**
+   * Performs a GET request to the CMS and returns the nested data at the provided `path`.
+   * @param {string} url CMS endpoint to fetch
+   * @param {Object} [config={}] axios.get request config
+   * @param {string|Array} [path="data"] lodash.get path to the nested data in the response
```
duncan-oxd: There are two approaches we can take here, but it could be confusing if we mix them both:

**Option A: "path" is the path in the axios response**
If we use `data.data` as the default path, let's update this comment here. But then I'd expect passing an empty path to return the full axios `response` object. This gives us the flexibility to check metadata from the axios response, but I don't think we'll never need to do that. (To the get full CMS response, you'd pass in `"data"`)

**Option B: "path" is the path in the CMS response**
I was thinking we could use `response.data` as the base and `data` as the default path, unless there's something else besides the data we'd want to get from the axios response. (To get the full CMS response, you'd pass in `""`)

Either one works. I'd go for B unless we anticipate needing to check the axios response code or something. Just make sure the comments align with the logic 👍

frontend/src/hooks/useCms.js
```diff
@@ -58,12 +59,25 @@ export default function useCms() {
...
-      // Read the Axios response payload and strip the extra Strapi "data" wrapper if it exists.
-      return response.data?.data ?? response.data;
+      // If an empty path is provided, return the whole response data
+      if (path.length === 0) {
+        return response.data;
+      }
+
+      // Default to Strapi's usual response.data.data shape,
+      // but allow callers to request a different path (e.g. response.data when meta is needed)
+      return get(response, path) ?? response.data;
```
duncan-oxd: `lodash.get` can take a third parameter for the default value to return if the `path` can't be resolved in the object. We could put `response.data` in there, but it might make sense to leave it out and let the function return undefined if the path is incorrect.

frontend/src/components/advisories/composite/dataTable/DataTable.jsx
```diff
@@ -513,6 +555,13 @@ DataTable.propTypes = {
     pageSize: PropTypes.number,
     pageSizeOptions: PropTypes.arrayOf(PropTypes.number),
     search: PropTypes.bool,
+    serverSide: PropTypes.bool,
+    totalItems: PropTypes.number,
+    currentPage: PropTypes.number,
+    onPageChange: PropTypes.func,
+    onPageSizeChange: PropTypes.func,
+    onSortChange: PropTypes.func,
+    onFilterChange: PropTypes.func,
```
duncan-oxd, line 564: Do we ever do anything special with the `options` object? It looks like these things inside it could all just be regular props

frontend/src/components/advisories/composite/dataTable/DataTable.jsx
```diff
@@ -292,18 +321,26 @@ export default function DataTable(props) {
 
     const columnId = getColumnId(column, index);
 
-    setSortConfig((currentSortConfig) => {
-      if (!currentSortConfig || currentSortConfig.columnId !== columnId) {
-        return { columnId, direction: "asc" };
-      }
+    let nextSort;
```
duncan-oxd, line 324: let's add comments to handleSort as well, to clarify what `nextSort` is, etc. 👍

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -40,6 +47,7 @@ import { updatePublicAdvisories } from "@/lib/advisories/utils/AdvisoryDataUtil"
  * @param {any} [defaultValue=0] Default value to return if the filter isn't found in storage
  * @returns {any} The filter value, or default if not found
  */
+
```
duncan-oxd, line 50:
```suggestion
```

An extra line was left between the jsdoc and the function 👀

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -453,26 +241,187 @@ export default function AdvisoryDashboard() {
...
+          },
+          { encodeValuesOnly: true },
+        );
+
+        const result = await cmsGet(
+          `/public-advisory-audits?${query}`,
+          {},
+          "data",
+        );
+        const rows = result.data ?? [];
```
duncan-oxd: This request is using the "data" path but then the next line is using the "data" path within that 👀 That could be combined to just get the axios `response.data.data` as usual from `cmsGet`

duncan-oxd (changes requested):
Nice! A couple of questions and small suggestions. Sorry if I'm confusing things with my cmsGet comments 😆 I hope it makes sense. Let me know if you have any questions

duncan-oxd (approved):
Great work!

PR #526, opened 2026-04-22
CMS-1668: Update Advisory Form fields, labels

duncan-oxd:
### Jira Ticket

CMS-1668

### Description

Updates to the Advisory form fields and labels. Mostly updating text and re-ordering a few fields and sections.

One of the items requires moving the red "*" for required fields, so I just used the global "append-required" class we made for DOOT and removed the old one. They're similar, but the DOOT one doesn't do anything specific with positioning so it's more adaptable.

Note that the Figma mockup says:

> More consideration is needed for BE updates 

This branch changes the _label_ for the "Description" field to say "Custom message" but it doesn't rename the field itself. That will happen in the future when they schedule backend changes so the frontend field can match the name of the backend field.

### DIFF
--- a/frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
+++ b/frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
@@ -1,6 +1,6 @@
 import { useState, useEffect } from "react";
 import PropTypes from "prop-types";
-import "./AdvisoryForm.css";
+import "./AdvisoryForm.scss";
 import { Button } from "@/components/advisories/shared/button/Button";
 import ButtonGroup from "react-bootstrap/ButtonGroup";
 import Btn from "react-bootstrap/Button";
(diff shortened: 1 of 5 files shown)


frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
```diff
@@ -1,14 +1,18 @@
 import { useState, useEffect } from "react";
 import PropTypes from "prop-types";
-import "./AdvisoryForm.css";
+import "./AdvisoryForm.scss";
```
duncan-oxd, line 3: Technically we can uses nested selectors in regular CSS now, but I renamed it to scss for consistency with the other pages. Even though there's not (currently) any special SCSS stuff in there.

frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.scss
```diff
@@ -1,3 +1,23 @@
...
+    margin: 32px 0;
+    padding: 4px 16px;
+  }
+
+  // Read-only field: use "disabled" background color
+  .standard-message-preview {
+    background: var(--surface-color-primary-disabled);
+  }
+}
+
```
duncan-oxd, line 20: Scoping all the new styles in this file to just the advisory form component. Some of the other styles are used on the ParkSearch page, and we can organize that later when we update that page.

ayumi-oxd (approved):
Looks good!

PR #536, opened 2026-04-29
CMS-1687: Add rec resource picker to advisory form

ayumi-oxd:
### Jira Ticket

CMS-1687

### Description

- Added the rec resource picker to the advisory form so that users can type and search a resource by name or ID and select it
- Added resource type icons


frontend/src/router/pages/advisories/advisory/Advisory.jsx
```diff
@@ -366,6 +372,20 @@ export default function Advisory({ mode }) {
...
+            if (recreationResourceInfo) {
+              const selRecreationResources = [];
+
+              recreationResourceInfo.forEach((resource) => {
+                selRecreationResources.push(
+                  recreationResources.find(
+                    (option) => option.value === resource.documentId,
+                  ),
+                );
+              });
```
duncan-oxd: It looks like this could be a `.map` instead needing a `.forEach` with a `.push` inside

frontend/src/components/advisories/composite/advisoryAreaPicker/RecreationResourcePicker.scss
duncan-oxd, line 1: Can we wrap the component and all the styles in a class name like `.recreation-resource-picker` to keep these styles scoped?

duncan-oxd (approved):
Code changes look good, and the new dropdown is cool. I left a few notes about tiny changes. Don't forget comments! 😅

PR #542, opened 2026-04-30
CMS-1700: Add action button to dashboard

ayumi-oxd:
### Jira Ticket

CMS-1700

### Description

- Added the action button to the advisory dashboard
- Added a new feature to unpublish an advisory with one click
- Displayed the flash message when unpublishing is succeed/failed


frontend/src/components/advisories/shared/tableActionButton/TableActionButton.scss
```diff
@@ -0,0 +1,52 @@
+.action-button {
+  &__dropdown-button {
+    .btn.dropdown-toggle {
+      background-color: var(--surface-color-background-white);
+      border: 1px solid var(--color-blue);
+      color: var(--color-blue);
+
+      &::after {
+        display: none;
+      }
```
duncan-oxd, line 11: Can you add a comment to say what this is doing?

frontend/src/components/advisories/shared/tableActionButton/TableActionButton.jsx
```diff
@@ -0,0 +1,72 @@
...
+  onEdit: PropTypes.func,
+  onUnpublish: PropTypes.func,
+};
+
+ActionButton.defaultProps = {
+  canUnpublish: false,
+  onView() {},
+  onEdit() {},
+  onUnpublish() {},
+};
```
duncan-oxd, line 65: The React 19 upgrade guide says they've [deprecated PropTypes](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops) in function components.

Our linter rules still want us to include propTypes because the project doesn't use typescript, but instead of defaultProps, we should use the default parameters in the function definition:

```js
// Something like this...
export function ActionButton({
  rowId,
  canUnpublish = false,
  onView = () => {},
  onEdit = () => {},
  onUnpublish = () => {},
}) {
```

We can also just make the functions required if it makes sense to do so

frontend/src/components/advisories/shared/actionButton/ActionButton.jsx
```diff
@@ -0,0 +1,72 @@
+import PropTypes from "prop-types";
+import Dropdown from "react-bootstrap/Dropdown";
+import DropdownButton from "react-bootstrap/DropdownButton";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faMagnifyingGlass, faChevronUp } from "@fa-kit/icons/classic/solid";
+import { faPen, faEyeSlash } from "@fa-kit/icons/classic/regular";
+
+import "./ActionButton.scss";
+
+export function ActionButton({
```
duncan-oxd: Small request: I created a component called ActionButtons for the forms in the 1.1.4 ticket 😅 I'll rename that to FormActionButtons or something, but maybe we can rename this to something like TableActionButton or RowActionButton just to keep it specific?

frontend/src/lib/advisories/utils/AdvisoryUnpublishPayload.js
```diff
@@ -0,0 +1,61 @@
+import moment from "moment";
+
+function mapDocumentIds(items) {
+  return (items || []).map((item) => item.documentId);
+}
```
duncan-oxd, line 10: Would this tweak improve readability a bit? 
```suggestion
function mapDocumentIds(items = []) {
  return items.map((item) => item.documentId);
}
```

(Thinking about how the function signature shows up in the editor)
Currently:

Image

vs. with the param default as an array:

Image

Or, even better with a jsdoc:

Image

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -69,6 +75,8 @@ function normalizePageFilterValues(filterValue) {
 
 export default function AdvisoryDashboard() {
   const { setError } = useContext(ErrorContext);
+  const auth = useAuth();
```
duncan-oxd, line 78: If you don't want to use auth.user.profile.name, we actually have a UserContext that provides `userDetails` and I don't think we use it anywhere in the app right now 😆 
It was intended to be used anywhere we need the logged in user's name (like in the top right corner or in the DOOT changelogs) but I don't know if we hooked it up anywhere.

frontend/src/router/pages/advisories/AdvisoryFlashMessage/AdvisoryFlashMessage.jsx
```diff
@@ -0,0 +1,27 @@
+import PropTypes from "prop-types";
+import FlashMessage from "@/components/FlashMessage";
+
+export default function AdvisoryFlashMessage({
```
duncan-oxd: It looks like the code is specific to the "unpublish" flash message - should we rename the file/component to reflect that? Or if it can be used for various kinds of Advisory things, we could rename the "unpublishFlashMessage" variables to something more general.

frontend/src/router/pages/advisories/AdvisoryFlashMessage/AdvisoryFlashMessage.jsx
```diff
@@ -0,0 +1,27 @@
...
+  unpublishFlashMessage,
+  closeUnpublishFlashMessage,
```
duncan-oxd: It makes sense to still call this prop "close" (or "onClose") 

If the other prop is an object that just expands to be props with the same names, we could use the spread syntax:
```jsx
<FlashMessage { ...flashMessageProps } onClose={onClose} />
```
But at that point, maybe we don't need this wrapper component at all, and we can pass the props directly to the regular FlashMessage component?

We use this pattern for the `ConfirmationMessage` in DOOT if you want a practical example. We _could_ use it for the FlashMessage in DOOT, but it looks like we haven't refactored that yet.

duncan-oxd (changes requested):
Nice, just a few questions and suggestions for small changes. Don't forget comments!

frontend/src/lib/advisories/utils/AdvisoryUnpublishPayload.js
ayumi-oxd: Thanks for the suggestions! I just learned that the default parameter only kicks in when the prop is undefined or has no value. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
However, advisory relation fields (e.g. sections, managementAreas) are often null rather than undefined, which could cause a type error with default parameters. So I believe the current code is the safer choice in this case 🙂

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
ayumi-oxd: It's just because `auth.user.profile.name` is already being used in several other places in ACT. It might be worth aligning ACT and DOOT to use the same context at some point. It'd be a nice cleanup task for the future 🙂

frontend/src/router/pages/advisories/AdvisoryFlashMessage/AdvisoryFlashMessage.jsx
ayumi-oxd: Good catch! This component wasn't needed 👍

frontend/src/lib/advisories/utils/AdvisoryUnpublishPayload.js
duncan-oxd: Ah, I didn't know it could be null. This is fine the way it is, in that case, but please update the jsdoc to indicate that the `items` param could be Array|null 👍 

Or: my top choice would be to make the function only accept an array, and then just change all the places that call the function (`mapDocumentIds(somevar || [])`). Then we wouldn't need to worry about cases where it's null or undefined 💡

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
duncan-oxd: It works either way, since the UserContext is available in the entire app already (anything that uses the MainLayout file is wrapped in it) but it's also good to be consistent across the ACT section at the moment, so this is fine ✅

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -31,6 +35,8 @@ import {
   buildFilter,
   buildSort,
 } from "@/lib/advisories/utils/AdvisoryDashboardQuery";
+import useAdvisoryUnpublish from "./useAdvisoryUnpublish";
+import useAdvisoryFlashMessage from "../AdvisoryFlashMessage/useAdvisoryFlashMessage";
```
duncan-oxd: Let's update to use the `@` alias for these import paths. I'm sure stuff will move around some day and we'll be doing lots of find-and-replace operations 😃

frontend/src/router/pages/advisories/advisoryDashboard/useAdvisoryUnpublish.js
```diff
@@ -0,0 +1,61 @@
+import { useCallback } from "react";
+
+import { ADVISORY_UNPUBLISH_QUERY } from "@/constants/advisoryQuery";
+import { buildUnpublishPayload } from "@/lib/advisories/utils/AdvisoryUnpublishPayload";
+
+export default function useAdvisoryUnpublish({
+  advisoryStatuses,
+  cmsGet,
+  cmsPut,
```
duncan-oxd: Hooks can actually import and use other hooks, so we can import useCms in this file, and then we'll have access to `cmsGet`, `cmsPut`, and `getAdvisoryStatuses` without needing to pass them in manually 👍 
and maybe even the flash message hook

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -970,6 +975,13 @@ export default function AdvisoryDashboard() {
           </div>
         </div>
       }
+      <FlashMessage
+        title={unpublishFlashMessage.title}
+        message={unpublishFlashMessage.message}
+        isVisible={unpublishFlashMessage.isVisible}
+        onClose={closeUnpublishFlashMessage}
+        variant={unpublishFlashMessage.variant}
+      />
```
duncan-oxd, line 984: I also remembered there's a global flash message in the root layout we can use:
```js
import globalFlashMessageContext from "@/contexts/FlashMessageContext";
// ...
const flashMessage = useContext(globalFlashMessageContext);
// ...
flashMessage.open(
  "Dates approved",
  `${name} ${season.operatingYear} dates marked as approved`,
);
```
Check out `FormPanel.jsx` in DOOT for an example. You get the same `open` and `close` functions, but you don't have to add a new FlashMessage component because it uses the one in the MainLayout. (The benefit is it will make sure all flash messages go through the same component to prevent a bunch of them stacking in the same place)

...but only two pages are using it so far 😆 Sorry for not mentioning that earlier. I think I lead you astray by showing you an older example on the Publishing page. Let's leave it for now and we can refactor all the flash messages some time in the future. Just letting you know for future reference, since I remembered it today 😅

duncan-oxd (approved):
Looks good! Just a request to update a jsdoc comment and some import lines, then some other FYI stuff you can ignore. I will use this unpublish function in my branch - updating the Advisory summary page to add an Unpublish button in the corner

PR #548, opened 2026-05-05
CMS-1688: Add plus font awesome icon

ayumi-oxd:
### Jira Ticket

CMS-1688

### Description

- Added a FontAwesome plus icon to the Create advisory button

### DIFF
--- a/frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.scss
+++ b/frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.scss
@@ -134,5 +134,12 @@
       color: var(--typography-color-primary) !important;
       font-weight: 400;
     }
+
+    button.bcgov-button {
+      .plus-icon {
+        width: 1rem;
+        height: 1rem;
+      }
+    }
   }
 }
--- a/frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.scss
+++ b/frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.scss
@@ -27,11 +27,6 @@
     font-weight: bold;
   }
 
-  svg {
-    width: 1.25rem;
-    height: 1.25rem;
-  }
-
   .revisionIcon {
     color: gray;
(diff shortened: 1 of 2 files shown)


frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.scss
```diff
@@ -134,5 +134,12 @@
       color: var(--typography-color-primary) !important;
       font-weight: 400;
     }
+
+    button.bcgov-button {
+      .plus-icon {
+        width: 1rem;
+        height: 1rem;
+      }
```
duncan-oxd: There's a very broad CSS selector that affects every SVG on this page and makes it 1.25rem in size:
```
.advisory-dashboard-page-wrap svg {...}
```
If we update that to make it more specific, then we won't need to add specific sizing for the other icons here. It'll make this plus icon 1em (which is 16px, the same as 1rem in this button)

duncan-oxd (changes requested):
I had a thought about fixing up the CSS selectors on this page, but if it's out of scope for this fix then I'd also be fine with merging now. Let me know what you think!

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.scss
ayumi-oxd: Good catch! This `svg {...}` isn't needed for now since we've changed the status icon to plain text. However, once we switch it back to an icon in the future, we'll need to revisit and add the icon-specific styling at that point.

duncan-oxd: That style was affecting the icons in the table column headers:
image

image

I don't mind the look now, but if we want to keep them at that larger size, we should write a more specific selector to target just the table header.

Everything else looks great, though! I'll approve it and you can discuss with UX if there's a preference about the icon sizes

duncan-oxd (approved):
One note about the icons in the table header, in case that needs a tweak. Ready to merge if not!

PR #523, opened 2026-04-20
CMS-1665: Add recreation district filter

ayumi-oxd:
### Jira Ticket

CMS-1665

### Description

- Added the RST district filter to the Advisory dashboard
- Updated the filters to the multi-select filters
- Added a reusable MultiSelect component for advisories so we don’t repeat the same code in multiple places
- Styled dropdown options to look like checkbox items and showed selected counts in the field label text

PR #551, opened 2026-05-06
CMS-1711: Add relational filters

ayumi-oxd:
### Jira Ticket

CMS-1711

### Description

- Region filter now matches advisories with selected `regions` or with `protectedAreas -> managementAreas -> region`
- District filter now matches advisories with selected `recreationDistricts` or with `recreationResources -> recreationDistrict`
- Updated the Clear filters button styling

### DIFF
--- a/frontend/src/lib/advisories/utils/AdvisoryDashboardQuery.js
+++ b/frontend/src/lib/advisories/utils/AdvisoryDashboardQuery.js
@@ -97,14 +97,9 @@ export function buildFilter(
 
   if (selectedDistrictIds.length > 0) {
     filters.push({
-      $or: [
-        { recreationDistricts: { documentId: { $in: selectedDistrictIds } } },
-        {
-          recreationResources: {
-            recreationDistrict: { documentId: { $in: selectedDistrictIds } },
-          },
-        },
-      ],
+      recreationResources: {
+        recreationDistrict: { documentId: { $in: selectedDistrictIds } },
+      },
     });
   }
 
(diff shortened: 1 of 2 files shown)


frontend/src/lib/advisories/utils/AdvisoryDashboardQuery.js
```diff
@@ -82,15 +82,29 @@ export function buildFilter(
...
+      ],
     });
   }
 
   if (selectedDistrictIds.length > 0) {
     filters.push({
-      recreationResources: {
-        recreationDistrict: { documentId: { $in: selectedDistrictIds } },
-      },
+      $or: [
```
molund: Was there a specific bug you were trying to fix with this change?  I think the second check is redundant and it behaves differently than the other filters

ayumi-oxd: It's a bug ticket https://bcparksdigital.atlassian.net/browse/CMS-1711.

- Issue: For example, selecting “Kootenay Okanagan Region” returns only advisories associated directly with that region.
- Expected: For example, selecting “Kootenay Okanagan Region” returns advisories that may not explicitly list the region but do include parks within it.

And the same thing applies to the RST districts filter.
The second check was meant to return all closures associated with resources in that district, even if a closure doesn't have a direct relation with a `recreationDistrict`.

molund: I added a commit to your CMS-1665 PR on April 28th that should have already fixed this for recreationDistricts. It looks like the ticket was opened on April 22.
https://github.com/bcgov/bcparks-staff-portal/pull/523/commits

ayumi-oxd: @molund Thanks! I completely forgot about that 😅 Updated it 👍

molund (approved):

PR #554, opened 2026-05-07
CMS-1685: Implement hasDates in Parks table

molund:
### Jira Ticket

CMS-1685

### Description

This change implements the new `hasDates` field in the UI table that lists all the parks in DOOT (not the database table called Parks which was originally a source of confusion) 

- Filtered out Parks where `Park.hasDates === false` from the main query in `backend/routes/api/parks.js`
- Updated all backend queries that were checking for `Feature.active === true` to also check for 
  `Feature.hasDates === true`.  `Feature.hasDates` has the exact same rules as `Feature.active` (features with `hasDates === false` are hidden from the UI)
- Updates the `create-test-park.js` script to set `hasDates: true` at the Park and Feature level.


ayumi-oxd (approved):
LGTM 👏

PR #555, opened 2026-05-07
CMS-1686: Implement hasDates in create-seasons script

molund:
### Jira Ticket

CMS-1686

### Description
This change adds checks for `Feature.hasDates` everywhere in the `create-seasons.js` script that we were previously checking for `Feature.active`. 

Since `active` and `hasDates` have the same behaviour, this approach should implement the new feature with minimal risk of introducing bugs.


ayumi-oxd (approved):
Looks good!

PR #559, opened 2026-05-11
CMS-1419: Clean up strapi-sync folder and tasks - Part 1

molund:
### Jira Ticket

CMS-1419

### Description
This is being split up into multiple PRs to make the code reviews more manageable.

The following requirements are implemented in this PR:  

(1) There’s a block of code in this file that imports Sections and Management Areas. Move that code into a separate task and include result counts similar to other scripts like tasks/import-date-types.

(2) Change the implemenation of the script so it can be imported or run stand-alone like other scripts, including transactions,

(3) There is also a cron-task script in the package.json file that should be updated to call this script using its new stand-alone implementation.  

(5) Change the cron task to run at 2am instead of 10pm. The park name data register script runs at 12:15am and this should be afterward

(6) Add a README to the strapi-sync folder

(7) Possible folder restructuring: move all scripts that get data from Strapi from backend/tasks to backend/strapi-sync/

(8) Add the missing import log events for management-areas and sections


backend/routes/api/parks.js
```diff
@@ -317,6 +320,13 @@ router.get(
             },
             // Publishable Seasons for the ParkArea
             seasonModel(currentYear),
+            // ParkAreaType for the ParkArea
+            {
+              model: ParkAreaType,
+              as: "parkAreaType",
+              attributes: ["id", "parkAreaTypeNumber", "name"],
+              required: true,
+            },
```
ayumi-oxd, line 329: I see some changes are the same as the ones in the alpha branch: https://github.com/bcgov/bcparks-staff-portal/pull/540/changes. Are these duplicates?

backend/utils/strapiApi.js
```diff
@@ -95,7 +81,7 @@ export async function getAllPages(endpoint, params = {}, delay = 1000) {
       hasMore = page < pageCount;
       currentPage++;
 
-      console.log(`Fetched page ${page}/${pageCount} from ${endpoint}`);
+      // console.log(`Fetched page ${page}/${pageCount} from ${endpoint}`);
```
ayumi-oxd: We can delete it if it's not needed anymore

ayumi-oxd (commented):
Looks good, just left a couple of questions 🙂

backend/routes/api/parks.js
molund: I think all the branch shuffling and force pushing for the prod deployment messed up the history.  I squashed my changes and it looks better now.

backend/utils/strapiApi.js
molund: Good call. I deleted it

ayumi-oxd (approved):
LGTM 👍

PR #562, opened 2026-05-13
CMS-1737: Update sort icon

ayumi-oxd:
### Jira Ticket

CMS-1737

### Description

- Update a sort icon
- Added advisory status sorting
- Polished resource sorting since it could display more than the resource/protected area name (region name, section name, etc ...)

### DIFF
--- a/frontend/src/components/advisories/composite/dataTable/DataTable.scss
+++ b/frontend/src/components/advisories/composite/dataTable/DataTable.scss
@@ -131,6 +131,12 @@
   font-size: 0.875rem;
   font-weight: inherit;
   text-align: left;
+
+  .warning-icon {
+    width: 1.25rem;
+    height: 1.25rem;
+  }
+
 }
 
 .data-table-header-button:focus,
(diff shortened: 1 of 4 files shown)


frontend/src/components/advisories/composite/dataTable/DataTable.scss
```diff
@@ -140,19 +145,29 @@
...
-}
-
-.data-table-header-button:hover .data-table-sort-indicator,
-.data-table-header-button:focus .data-table-sort-indicator,
-.data-table-header-button:focus-visible .data-table-sort-indicator,
-.data-table-sort-indicator.is-active {
-  opacity: 1;
+  color: var(--icons-color-disabled);
+  display: inline-flex;
+  flex-direction: column;
```
duncan-oxd: Would it be easier to use the combined "sort" icon with both arrows when the sort state isn't set?
https://fontawesome.com/icons/classic/solid/sort

frontend/src/components/advisories/composite/dataTable/DataTable.scss
```diff
@@ -131,6 +131,11 @@
   font-size: 0.875rem;
   font-weight: inherit;
   text-align: left;
+
+  .warning-icon {
+    width: 1.25rem;
+    height: 1.25rem;
```
duncan-oxd: FontAwesome icons all have the svg `viewbox` attribute set so they'll always be a square. I think we only need to set the `height` (or the `font-size`) to size them. If they ever show up as a non-square, we could put a global rule somewhere to give all `.fa` icons an aspect-ratio of 1 or something.

Try changing this to `font-size: 1.25rem` and it should end up the same size and shape unless something else on the page is affecting it.

duncan-oxd (changes requested):
I suggested a few small CSS tweaks. Also note that I changed the "baseFilters" section in #570 so whoever merges second will need to be aware of a small possible conflict there.

frontend/src/components/advisories/composite/dataTable/DataTable.scss
ayumi-oxd: When sorting is active, either the up or down icon is hidden to indicate the selected order (asc or desc). Because of this, it was easier to use separate sort icons rather than a combined one.

duncan-oxd: I meant we could use all three icons: keep sort-up and sort-down, but when sorting is disabled, just use "sort" to show both arrows so we don't need special CSS to position the two icons to get the same effect
imageimage

(They're aligned how we want already so we can just swap out the icon name)

ayumi-oxd: That's a good point! Updated it.

duncan-oxd (approved):
Nice! Great job simplifying it with that third icon 👍

PR #564, opened 2026-05-13
CMS-1699: Update filter close button

ayumi-oxd:
### Jira Ticket

CMS-1699

### Description

- Update filter close button colour (opacity)


duncan-oxd (approved):
LGTM! If we want to override the Bootstrap defaults _globally_ (in case we use this chip component somewhere else in the future) we can also override the `--bs-btn-close-opacity` variable (defaults to 0.5). Tweaking the Bootstrap theme to align with the design system is a larger effort for the future, though 🤔

PR #567, opened 2026-05-15
CMS-1739: Update action button styles & behavior

duncan-oxd:
### Jira Ticket

CMS-1739

### Description

From Figma screen 1.0.1, items 13, 14, and 17:
Updated the "Actions" dropdown to just show the chevron icon and have the text hidden.
Updated the popper js options to keep the menu inside the viewport if you scroll away from the toggle button. I had to change the component itself to use the base Dropdown component instead of DropdownButton to use the popperConfig prop
Updated the icon and some minor sizing/style stuff to have the column take up less horizontal space without looking cramped


ayumi-oxd (approved):
LGTM 👏

PR #571, opened 2026-05-19
CMS-1732: Add mark reviewed

ayumi-oxd:
### Jira Ticket

CMS-1732

### Description

- Added "Mark reviewed" option in the action dropdown
  - When it's clicked, it updates the public advisory audit with `reviewedAt` and `reviewedByName`
  - The advisory will NOT appear in the review tab once it's reviewed 
  - It keeps the advisory status, but the status `HQR` will be updated to `PUB` or `SCH` based on the advisory date
- Added `tabParam` to define from which tab page was accessed so that users can go back to the previous location (tab)
- Added a flash message to show if marking the advisory as reviewed is successful

### DIFF
--- a/frontend/src/hooks/advisories/useAdvisoryMarkReviewed.js
+++ b/frontend/src/hooks/advisories/useAdvisoryMarkReviewed.js
@@ -74,7 +74,6 @@ export default function useAdvisoryMarkReviewed({
           data: buildReviewPayload(
             advisoryData,
             reviewedStatus.documentId,
-            reviewedStatus.code,
             reviewedByName,
             isApprover ? "approver" : "submitter",
           ),
--- a/frontend/src/hooks/advisories/useAdvisoryMarkReviewed.js
+++ b/frontend/src/hooks/advisories/useAdvisoryMarkReviewed.js
@@ -47,7 +47,6 @@ function resolveReviewedStatus(rowData, advisoryStatuses) {
 export default function useAdvisoryMarkReviewed({
   advisoryStatuses,
   reviewedByName,
-  isApprover,
   openMarkReviewedError,
   openMarkReviewedSuccess,
   onSuccess,
@@ -75,7 +74,6 @@ export default function useAdvisoryMarkReviewed({
             advisoryData,
             reviewedStatus.documentId,
             reviewedByName,
-            isApprover ? "approver" : "submitter",
           ),
         });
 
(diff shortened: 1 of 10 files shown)


frontend/src/hooks/advisories/useAdvisoryMarkReviewed.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // If the status is currently UNP, keep it UNP but just update the reviewedAt/reviewedBy fields
+  if (statusCode === "UNP") {
+    return unpublishedStatus;
+  }
+
+  // If the status is currently HQR, set to SCH if future posting date, else PUB
+  if (statusCode === "HQR") {
+    return moment(rowData.advisoryDate).isAfter(moment())
```
duncan-oxd: If we're hoping to eventually port all the `moment` code over to `date-fns`, we could use date-fns here since it's new code being added. I think something like this would work:
```js
return isFuture(rowData.advisoryDate) ? scheduledStatus : publishedStatus;
```

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -133,6 +133,28 @@ export default function AdvisoryDashboard({
...
+          variant: "error",
+        },
+      );
+    },
+    [globalFlashMessage],
+  );
+
+  const openMarkReviewedSuccess = useCallback(
+    (message) => {
+      globalFlashMessage.open("Marked Advisory / Closure reviewed", message, {
```
duncan-oxd: We might need to confirm the capitalization in this message, I think the latest requirements are favoring sentence case for stuff like this

duncan-oxd (changes requested):
A few small things to consider, but otherwise it looks good! 
In the future, we might decide to use localStorage/sessionStorage to remember what Advisory tab view the user was on but the tab parameter works just as well 👍

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
molund: @ayumi-oxd I would check with UX on the correct capitalization, but it should be either "Advisory / Closure" or "advisory / closure", not "Advisory / closure".  I think "advisory / closure" is most likely correct.

molund (approved):

duncan-oxd (approved):

PR #572, opened 2026-05-20
CMS-1683: Rename and remove css classes beginning with "ad-"

molund:
### Jira Ticket

CMS-1683

### Description

Some aggressive ad-blockers may hide parts of the Advisories portal UI because some elements are named “ad-text”, “ad-button” etc.

This PR changes the prefix on CSS classes from  `ad-` to `act-` and removes CSS classes that were not referenced in code or markup.

### DIFF
--- a/frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
+++ b/frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
@@ -362,7 +362,7 @@ export default function AdvisoryForm({
 
     return (
       <div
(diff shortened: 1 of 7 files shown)


ayumi-oxd (approved):

ayumi-oxd:
I haven't tested it with any ad blocker tools, but I wonder if they might block `adv-(advertisement)` too. Hopefully not 😅

molund:
> I haven't tested it with any ad blocker tools, but I wonder if they might block `adv-(advertisement)` too. Hopefully not 😅

Good point. I didn't want to use "advisory-" because there were conflicts with some existing classes, so I changed it to "act-"

PR #573, opened 2026-05-21
CMS-1735: Add program area selector

ayumi-oxd:
### Jira Ticket

CMS-1735

### Description

- Added the program area selector
- Added a filter to check advisories' relation with protectedAreas/resources

### DIFF
--- a/frontend/src/components/advisories/shared/multiSelect/MultiSelect.scss
+++ b/frontend/src/components/advisories/shared/multiSelect/MultiSelect.scss
@@ -3,11 +3,3 @@
   align-items: center;
   gap: 8px;
 }
-
-.bcgov-select {
-  &.bcgov-select--has-selection {
-    [class*="-placeholder"] {
-      color: var(--typography-color-primary) !important;
-    }
-  }
-}
--- a/frontend/src/components/advisories/shared/singleSelect/SingleSelect.jsx
+++ b/frontend/src/components/advisories/shared/singleSelect/SingleSelect.jsx
@@ -4,7 +4,13 @@ import PropTypes from "prop-types";
 import Form from "react-bootstrap/Form";
 import Select from "react-select";
 
-export function SingleSelect({ label, value, options, onChange, placeholder }) {
+export function SingleSelect({
+  label = "",
+  value = null,
+  options = [],
+  onChange = () => {},
+  placeholder = "Select...",
+}) {
   const generatedId = useId();
   const hasSelection = Boolean(value);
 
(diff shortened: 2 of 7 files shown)


frontend/src/components/advisories/shared/multiSelect/MultiSelect.scss
```diff
@@ -3,11 +3,3 @@
...
   gap: 8px;
 }
-
-.bcgov-select {
-  &.bcgov-select--has-selection {
-    [class*="-placeholder"] {
-      color: var(--typography-color-primary) !important;
-    }
-  }
-}
```
duncan-oxd, line 13: Did this move to the page.scss file to keep it scoped to elements with the "advisories-styles" class? it looks like they would still work here, but do they interfere with something in DOOT? (I notice we have two files called multiselect.scss)

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -287,6 +294,49 @@ export default function AdvisoryDashboard({
     );
   }
 
+  function handleProgramAreaChange(option) {
```
duncan-oxd, line 330: It's always good to include some brief comments with function definitions, especially if the name is just handle<event name>

It's a good habit to get into, and I think I need to write a skill file for my Copilot to do the same 😁

frontend/src/components/advisories/shared/singleSelect/SingleSelect.jsx
```diff
@@ -0,0 +1,51 @@
...
+  placeholder: PropTypes.string,
+};
+
+SingleSelect.defaultProps = {
+  label: "",
+  value: null,
+  options: [],
+  onChange() {},
+  placeholder: "Select...",
+};
```
duncan-oxd: It's good to be consistent with the existing select component, but since this is a new file we're adding and defaultProps has been deprecated, I think we should remove this block and put these as the default parameter values on line 7

duncan-oxd (changes requested):
A few small questions and one tiny change request. It's kind of a coin-flip for some of these style choices because it's awkward to have half-half old and new code with two different styles, but I also don't to build more and more on top of the old code if we get a chance to update it 😅

frontend/src/components/advisories/shared/multiSelect/MultiSelect.scss
ayumi-oxd: It doesn't since this style is wrapped with `.advisories-styles` in `page.scss`, and I don't see `.bcgov-select` is used in DOOT

frontend/src/components/advisories/shared/singleSelect/SingleSelect.jsx
ayumi-oxd: Sorry I keep forgetting defaultProps is deprecated 😅

duncan-oxd (approved):

PR #575, opened 2026-05-21
CMS-1716: Fix for tooltip icons not keyboard focusable

molund:
### Jira Ticket

CMS-1716

### Description
This change allows tooltip icons to be focused using the keyboard.

### DIFF
--- a/frontend/src/components/TooltipWrapper.jsx
+++ b/frontend/src/components/TooltipWrapper.jsx
@@ -13,13 +13,18 @@ export default function TooltipWrapper({
 
   return (
     <>
-      <span
+      <button
         ref={target}
+        type="button"
         onMouseEnter={() => setShow(true)}
         onMouseLeave={() => setShow(false)}
+        onFocus={() => setShow(true)}
+        onBlur={() => setShow(false)}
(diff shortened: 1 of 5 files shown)


duncan-oxd (approved):
Looks good to me. Some day we'll consolidate our two different tooltip approaches between DOOT and ACT 😆 

We could add a css rule or a bootstrap class to force the regular mouse cursor instead of the 👆 hand cursor, if it incorrectly suggests "this is something to click on." I think Bootstrap gives that style to all button elements. Maybe it just means "this something the mouse interacts with" ... open to interpretation!

molund:
> Looks good to me. Some day we'll consolidate our two different tooltip approaches between DOOT and ACT 😆
> 
> We could add a css rule or a bootstrap class to force the regular mouse cursor instead of the 👆 hand cursor, if it incorrectly suggests "this is something to click on." I think Bootstrap gives that style to all button elements. Maybe it just means "this something the mouse interacts with" ... open to interpretation!

Copilot actually leans toward keeping the pointer, since the icon does react (hover/focus shows the tooltip). Bootstrap examples follow the same pattern, so I’d lean toward keeping it.

PR #576, opened 2026-05-21
CMS-1738: Add tooltip to all count badge

ayumi-oxd:
### Jira Ticket

CMS-1738

### Description

- Added tooltip to all count badge
- Created the count badge component so it can be reused in the dashboard

### DIFF
--- a/frontend/src/components/advisories/shared/countBadge/CountBadge.jsx
+++ b/frontend/src/components/advisories/shared/countBadge/CountBadge.jsx
@@ -1,15 +1,13 @@
+import { useId } from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import Badge from "react-bootstrap/Badge";
 import OverlayTrigger from "react-bootstrap/OverlayTrigger";
 import Tooltip from "react-bootstrap/Tooltip";
 
-import "./CountBadge.scss";
-
-export function CountBadge({ index, label, documentId, title, tooltipText }) {
-  // One advisory can have multiple count badges
-  // e.g. "count-badge-tooltip-123abc--0" and "count-badge-tooltip-123abc--1"
-  const tooltipId = `count-badge-tooltip-${documentId}-${index}`;
+export function CountBadge({ label, documentId, title, tooltipText }) {
+  const generatedId = useId();
+  const tooltipId = `count-badge-tooltip-${documentId}-${generatedId.replace(/[:]/gu, "")}`;
   const ariaLabelText = `Open advisory summary for ${title}`;
 
   return (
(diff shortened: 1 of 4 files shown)


frontend/src/components/advisories/shared/countBadge/CountBadge.jsx
```diff
@@ -0,0 +1,37 @@
+import { useId } from "react";
+import PropTypes from "prop-types";
+import { Link } from "react-router-dom";
+import Badge from "react-bootstrap/Badge";
+import OverlayTrigger from "react-bootstrap/OverlayTrigger";
+import Tooltip from "react-bootstrap/Tooltip";
+
+export function CountBadge({ label, documentId, title, tooltipText }) {
+  const generatedId = useId();
+  const tooltipId = `count-badge-tooltip-${documentId}-${generatedId.replace(/[:]/gu, "")}`;
```
duncan-oxd: if useId returns a unique ID every time we call it, we probably don't need to add anything for `tooltipId` here (unless it makes it easier to identify in the DOM)

duncan-oxd: The square brackets might be redundant in the regular expression, see if this works:

```suggestion
  const generatedId = useId();
  const tooltipId = `count-badge-tooltip-${documentId}-${generatedId.replace(/:/gu, "")}`;
```

frontend/src/components/advisories/shared/countBadge/CountBadge.jsx
```diff
@@ -0,0 +1,37 @@
...
+  label: PropTypes.string.isRequired,
+  documentId: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  tooltipText: PropTypes.string,
+};
+
+CountBadge.defaultProps = {
+  title: "",
+  tooltipText: "",
+};
```
duncan-oxd: I had a similar thought in another PR, let's remove the defaultProps from any new files and instead just use the default parameter values in the function definition on line 8.
Less refactoring work for us to do later 👍

duncan-oxd (changes requested):
Nice one, a small change about the propTypes and some stuff to consider with the useId hook

molund (approved):

frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -94,6 +94,15 @@ function ReviewEmptyState() {
   );
 }
 
+// Format the tooltip text for count badges that indicate additional associated resources
+function formatCountBadge(
```
duncan-oxd, line 98: Maybe this should be `formatCountBadgeTooltip` but it makes sense either way 👍

duncan-oxd (approved):
Looks good! I didn't know we had to provide a unique ID for the react-bootstrap tooltips? That's inconvenient!

molund (approved):

duncan-oxd (approved):

PR #582, opened 2026-05-25
CMS-1730: Add review tab icons

ayumi-oxd:
### Jira Ticket

CMS-1730

### Description

- Added the advisory review status icons that visually highlight advisories that are new, updated, unpublished, expiring, expired, or ended
- Added a centralized `REVIEW_STATUS` constant object to define all possible review statuses
- Implemented a `buildAdvisoryReviewStatuses` utility function to determine applicable review statuses for each advisory based on its metadata, and attached these statuses to each advisory record
- Enhanced review dashboard query logic to better filter advisories based on their review status and activity, ensuring the dashboard reflects the new review status system


molund (approved):
Looks good.  I think the "Warning" text should say "Ended" but it matches the design.  This will be a good one to include in the demo tomorrow.

PR #584, opened 2026-05-26
NOBUG: Fix pagination handling for large Strapi collections

molund:
### Jira Ticket

CMS-1687

### Description
This fixes a bug where only the first 2000 items were coming back from the Strapi recreation-resources collection, because Strapi is configured to limit single requests to 2000 items at a time.  

The fix borrows a pattern from the backend project which can hopefully be moved to a shared library at a later date. 

Also converted the dropdown to use `react-select/async` to improve preformance issues with 6500 items


ayumi-oxd (approved):
Looks good!

PR #591, opened 2026-05-28
CMS-1733: Updates to audit field usage

molund:
### Jira Ticket

CMS-1733

### Description
- Added logic to populate the createdByEmail field when a new advisory is created
- Added logic to set createdByRole and modifiedByRole to three values instead of two ("approver", "submitter", "contributor")


frontend/src/router/pages/advisories/advisory/Advisory.jsx
```diff
@@ -1026,9 +1028,10 @@ export default function Advisory({ mode }) {
         publishedAt: new Date(),
         isLatestRevision: true,
         createdByName: auth.user?.profile?.name,
+        createdByEmail: auth.user?.profile?.email,
         reviewedByName: null,
         reviewedDate: null,
-        createdByRole: isApprover ? "approver" : "submitter",
+        createdByRole: getUserAdvisoryRole(),
```
ayumi-oxd, line 1034: I wonder if we still need `createdByRole`

ayumi-oxd (approved):
LGTM!

frontend/src/router/pages/advisories/advisory/Advisory.jsx
molund: I don't see any logic currenly attached to it, but it seems to work and it might be useful for debugging or troubleshooting.

PR #594, opened 2026-06-01
CMS-1764: Update advisory status to icon

ayumi-oxd:
### Jira Ticket

CMS-1764

### Description

- Update advisory status text to icon for approvers
- Display tooltip when icon is on hover

### DIFF
--- a/frontend/src/router/pages/advisories/advisorySummary/AdvisorySummary.jsx
+++ b/frontend/src/router/pages/advisories/advisorySummary/AdvisorySummary.jsx
@@ -471,6 +471,7 @@ export default function AdvisorySummary() {
                             Advisory #{advisory.advisoryNumber}
                             <StatusBadge
                               status={advisoryStatusCode}
+                              approver={canMarkReviewed}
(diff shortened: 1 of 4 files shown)


frontend/src/router/pages/advisories/advisorySummary/AdvisorySummary.jsx
molund: Maybe we should rename this to `isApprover` since it's being used for more than one purpose now?

ayumi-oxd: I'd like to hear @duncan-oxd 's opinion, since he originally introduced it 🙂

molund (approved):

frontend/src/router/pages/advisories/advisorySummary/AdvisorySummary.jsx
duncan-oxd: That's a good plan, `isApprover` is a clearer name for what's checking anyway. I think a previous version was checking multiple roles but it wound up only being the one 👍

PR #604, opened 2026-06-04
CMS-1804: Display both park and resource if an advisory has both

ayumi-oxd:
### Jira Ticket

CMS-1804

### Description

- Displayed both parks and resources in the Associated resources column if an advisory has both

### DIFF
--- a/frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
+++ b/frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
@@ -970,17 +970,17 @@ export default function AdvisoryDashboard({
             );
           }
 
-          // Display parks and recreation resources
-          const resources = rowData.recreationResources ?? [];
+          // Display parks and rec resources
+          const recResources = rowData.recreationResources ?? [];
           const parks = rowData.protectedAreas ?? [];
-          const resourcesCount = resources.length;
+          const recResourcesCount = recResources.length;
           const parksCount = parks.length;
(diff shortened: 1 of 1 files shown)


frontend/src/router/pages/advisories/advisoryDashboard/AdvisoryDashboard.jsx
```diff
@@ -970,65 +970,49 @@ export default function AdvisoryDashboard({
...
-          const displayedProtectedAreas = rowData.protectedAreas.slice(
-            0,
-            displayCount,
-          );
+          // Display parks and recreation resources
+          const resources = rowData.recreationResources ?? [];
+          const parks = rowData.protectedAreas ?? [];
+          const resourcesCount = resources.length;
+          const parksCount = parks.length;
+          const displayedResources = resources.slice(0, displayCount);
```
molund: Maybe we should call these "recResources" instead of just "resources" now that the UI is calling everything a resource?

molund:
Looks good.  I tested on my local and it added the +1 after the 7th resource was included. But when there are only parks it seems to add it after the 4th. I'm not sure if that's intended.

molund (approved):

ayumi-oxd:
> Looks good. I tested on my local and it added the +1 after the 7th resource was included. But when there are only parks it seems to add it after the 4th. I'm not sure if that's intended.

Good catch! It was counting recreation resources and parks separately. I've updated it to combine all resources first and then display the number of additional resources beyond the first three.

PR #607, opened 2026-06-05
CMS-1793: Update public access status select field

ayumi-oxd:
### Jira Ticket

CMS-1793

### Description

- Changed the name from Resource status to Public access status
- Added category in the dropdown
- Removed pre-select "Open" status for RST users

### DIFF
--- a/frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
+++ b/frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
@@ -41,6 +41,7 @@ import AdvisoryAreaPicker from "@/components/advisories/composite/advisoryAreaPi
 import CKEditor from "@/components/advisories/composite/ckeditor/CKEditor";
 import PrimaryActions from "@/components/advisories/composite/advisoryForm/PrimaryActions";
 import DraftButton from "@/components/advisories/composite/advisoryForm/DraftButton";
+import AccessStatusPicker from "@/components/advisories/composite/advisoryForm/AccessStatusPicker";
 
 export default function AdvisoryForm({
   mode,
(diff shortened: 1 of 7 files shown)


frontend/src/components/advisories/shared/categorySelect/CategorySelect.jsx
```diff
@@ -0,0 +1,111 @@
...
+    const groups = options.reduce((accumulator, option) => {
+      const category = option[categoryKey];
+
+      if (!accumulator[category]) {
+        accumulator[category] = [];
+      }
+
+      accumulator[category].push(option);
+      return accumulator;
+    }, Object.create(null));
```
duncan-oxd: If you want a one-liner for this, we've got lodash-es `groupBy(options, "category")` available here

frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
```diff
@@ -543,13 +545,13 @@ export default function AdvisoryForm({
...
-            id="resource-status"
-            options={accessStatuses}
-            value={accessStatuses.filter((e) => e.value === accessStatus)}
-            onChange={(e) => setAccessStatus(e ? e.value : 0)}
-            placeholder="Search or select resource status"
-            className="bcgov-select"
+          <AccessStatusPicker
+            accessStatus={accessStatus}
+            accessStatuses={accessStatuses}
+            setAccessStatus={setAccessStatus}
```
duncan-oxd: It probably doesn't matter and I don't know if we have an exact convention we're following, but maybe this prop should be called `onChange` or `onSelect`?

duncan-oxd (approved):
Looks good! The new selection control seems to work really well! 
I left a couple of notes to consider but it looks good to me!

frontend/src/components/advisories/composite/advisoryForm/AdvisoryForm.jsx
ayumi-oxd: It is used in `CategorySelect.jsx` 🙂

PR #617, opened 2026-06-12
CMS-1723: Update event type select field

ayumi-oxd:
### Jira Ticket

CMS-1723

### Description

- Updated the event type select field
- Added a option category in the dropdown

### Todo
- [x] Move duplicated functions in AccessStatusPicker and EventTypePicker to CategorySelect after CMS-1793


molund:
It works really well.  I noticed public access status is missing the Seasonal Restrictions option on my machine. I think we might have hidden it in the past, but it's shown in the wireframes?

ayumi-oxd:
> It works really well. I noticed public access status is missing the Seasonal Restrictions option on my machine. I think we might have hidden it in the past, but it's shown in the wireframes?

Yes it's shown in the wireframe, and it's shown on Alpha-DEV: https://alpha-dev-cms.bcparks.ca/admin/content-manager/collection-types/api::access-status.access-status/i7sga9a0kng5vehj612eycyi

molund (approved):

PR #620, opened 2026-06-15
CMS-1773: Add internal notes row to table

ayumi-oxd:
### Jira Ticket

CMS-1773

### Description

- Added an internal notes row to the dates table for approvers
- Users can expand the accordion to see the notes


backend/routes/api/parks.js
```diff
@@ -59,6 +60,18 @@ function seasonModel(minYear, required = true) {
           },
         ],
       },
+      {
```
molund, line 63: I thought we were just going to check if a note existed in this query and then load it with Javascript when someone clicks the link?

ayumi-oxd: As we discussed, that'll be a future work in CMS-1844

molund: I updated the Lazy Load ticket to make sure this is included in the scope.

molund:
Shouldn't this be merged to main since it's a DOOT ticket?

ayumi-oxd:
> Shouldn't this be merged to main since it's a DOOT ticket?

@molund As we discussed, changed the target branch to main

molund (approved):

PR #624, opened 2026-06-17
CMS-1776: Update standard message select field

ayumi-oxd:
### Jira Ticket

CMS-1776

### Description

- Updated the standard message select field
- Added a option category (`standardMessage.eventType.groupLabel`) in the dropdown


duncan-oxd (approved):
Looks like a nice improvement!

PR #625, opened 2026-06-19
CMS-1714: Change main-dev helm to use prod Keycloak

molund:
### Jira Ticket

CMS-1714

### Description
Updating the staff portal main-dev helm config to use the prod Keycloak instead of the dev Keycloak.
This was missed previously in https://github.com/bcgov/bcparks-staff-portal/pull/565

### DIFF
--- a/helm/deployment/values-dev.yaml
+++ b/helm/deployment/values-dev.yaml
@@ -1,5 +1,5 @@
 cluster:
-  ssoAuthUrl: https://dev.loginproxy.gov.bc.ca/auth
+  ssoAuthUrl: https://loginproxy.gov.bc.ca/auth
   cmsUrl: https://dev-cms.bcparks.ca
 
 frontend:


duncan-oxd (approved):
Discussing on Slack whether we want to ultimately switch the dev environment back to using the dev keycloak, or fully make it use prod keycloak. (Depends on what kind of testing is on the horizon for that environment)

If using prod keycloak is the current goal, then this is approved 👍

molund:
> Discussing on Slack whether we want to ultimately switch the dev environment back to using the dev keycloak, or fully make it use prod keycloak. (Depends on what kind of testing is on the horizon for that environment)
> 
> If using prod keycloak is the current goal, then this is approved 👍

It's not that simple to change it back. We used the production Keycloak for the training environment to avoid requiring Park Operators to create separate BCeID accounts for test and production when doing training and real DOOT work. This was never an issue for IDIR users, since they use the same identity across all Keycloak environments. The BCeID identity provider, however, maintains separate user stores for each environment.

Using production Keycloak for training created a downstream side effect: because DOOT and ACT share the same frontend, ACT was effectively broken in the training environment. This wasn’t an issue early on when users were directed to the `/dates/` subdirectory (https://training-staff.bcparks.ca/dates/). However, that safeguard is easily bypassed—users naturally navigate to the root URL (https://training-staff.bcparks.ca/), which exposes the shared frontend and results in errors.

PR #631, opened 2026-06-22
CMS-1613: Update frontend headings to show two year spans

duncan-oxd:
### Jira Ticket

CMS-1613

### Description

Updated the Form Panel heading and the table column headings for Features with "Dates can span 2 years" flags.
In these cases, we'll now show a date range ("2026 – 2027") instead of just the operating year ("2026")

To keep things simple in the template, I added the datesCanSpan2Years property to the season endpoint. Always false for parks, same as Feature for features, and calculated for Areas.

### DIFF
--- a/frontend/src/components/FormPanel.jsx
+++ b/frontend/src/components/FormPanel.jsx
@@ -152,7 +152,22 @@ function SeasonForm({
     previousWinter: previousWinterSeasonDates,
     ...seasonMetadata
   } = data || {};
-  const currentYear = season?.operatingYear;
+
+  // Derive the header text from the season data
+  const yearHeaderText = useMemo(() => {
+    // operatingYear is always defined for a Season, so data is still loading if it's undefined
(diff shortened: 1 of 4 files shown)


frontend/src/components/FormPanel.jsx
```diff
@@ -152,7 +152,22 @@ function SeasonForm({
...
+
+  // Derive the header text from the season data
+  const yearHeaderText = useMemo(() => {
+    // operatingYear is always defined for a Season, so data is still loading if it's undefined
+    if (!season?.operatingYear) return "";
+
+    const operatingYear = season.operatingYear;
+    const nextYear = operatingYear + 1;
+
+    if (season.datesCanSpan2Years) {
```
duncan-oxd, line 164: We will probably add a check for the winter fee flag to this block very soon

ayumi-oxd (approved):

PR #632, opened 2026-06-22
CMS-1811: Role checks for handling the advisory contributor role

molund:
### Jira Ticket

CMS-1811

### Description
Add role checks for handling the new advisory contributor role.

### Notes

This tickets implements UI elements of the contributor role, but for full usage comparable to DOOT more work is required, especially around implementing bundle access restrictions.  POs in DOOT only have access to parks in their bundle.  This is currently implemented to users in the contributor role can draft advisories for all BC Parks and RST resources. 

A Keycloak group has been created on the Dev keycloak environment called "BCP Park Operator ACT TESTING" but there will be no actual groups that only have the contributor role in the prod environment as part of the MVP launch. 

Also includes a fix for the UnsavedChangesDialog incorrectly appearing after a successful form save. The unsaved-changes flag was tracked with useState, which updates asynchronously — by the time navigate() fired after saving, the route blocker and beforeunload guard still read the stale true value. Switching to useRef ensures the flag is cleared synchronously before navigation is attempted.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -70,19 +70,18 @@ const apiRouter = express.Router();
 // Apply JWT check and user middleware to all /api routes
 apiRouter.use(checkJwt);
 
+// Ensure user has STAFF_PORTAL_USER role to access any /api route
+apiRouter.use(checkPermissions([ROLES.STAFF_PORTAL_USER]));
+
 // Attach user data from the DB to the request object
 apiRouter.use(usersMiddleware);
 
-apiRouter.use("/user", checkPermissions([ROLES.STAFF_PORTAL_USER]), userRoutes);
+apiRouter.use("/user", userRoutes);
 apiRouter.use("/parks", checkPermissions([ROLES.DOOT_USER]), parkRoutes);
 apiRouter.use("/seasons", checkPermissions([ROLES.DOOT_USER]), seasonRoutes);
 apiRouter.use("/export", checkPermissions([ROLES.DOOT_USER]), exportRoutes);
 apiRouter.use("/publish", checkPermissions([ROLES.DOOT_USER]), publishRoutes);
-apiRouter.use(
-  "/filter-options",
-  checkPermissions([ROLES.DOOT_USER]),
-  filterOptionsRoutes,
-);
+apiRouter.use("/filter-options", filterOptionsRoutes);
 
 app.use("/api", apiRouter);
 
--- a/backend/index.js
+++ b/backend/index.js
@@ -70,18 +70,19 @@ const apiRouter = express.Router();
 // Apply JWT check and user middleware to all /api routes
 apiRouter.use(checkJwt);
 
-// Ensure user has DOOT_USER role to access any /api route
-apiRouter.use(checkPermissions([ROLES.DOOT_USER]));
-
 // Attach user data from the DB to the request object
 apiRouter.use(usersMiddleware);
 
-apiRouter.use("/user", userRoutes);
-apiRouter.use("/parks", parkRoutes);
-apiRouter.use("/seasons", seasonRoutes);
-apiRouter.use("/export", exportRoutes);
-apiRouter.use("/publish", publishRoutes);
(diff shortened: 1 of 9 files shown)


backend/index.js
```diff
@@ -70,17 +70,17 @@ const apiRouter = express.Router();
...
 apiRouter.use("/user", userRoutes);
-apiRouter.use("/parks", parkRoutes);
-apiRouter.use("/seasons", seasonRoutes);
-apiRouter.use("/export", exportRoutes);
-apiRouter.use("/publish", publishRoutes);
+apiRouter.use("/parks", checkPermissions([ROLES.DOOT_USER]), parkRoutes);
+apiRouter.use("/seasons", checkPermissions([ROLES.DOOT_USER]), seasonRoutes);
+apiRouter.use("/export", checkPermissions([ROLES.DOOT_USER]), exportRoutes);
+apiRouter.use("/publish", checkPermissions([ROLES.DOOT_USER]), publishRoutes);
 apiRouter.use("/filter-options", filterOptionsRoutes);
```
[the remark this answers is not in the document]
molund: This is fine. filter-options is read-only and it doesn't expose any private data. This endpoint may be used for advisories too at some point in the future.

frontend/src/hooks/useNavigationGuard.js
```diff
@@ -1,12 +1,12 @@
...
-// Prevent document-level unloads when hasChanges is true.
+// Prevent document-level unloads when hasChanges() returns true.
 // This hook handles native browser prompts for reload/tab close/external document navigation.
 
 export default function useNavigationGuard(hasChanges) {
   useEffect(() => {
     function handleBeforeUnload(event) {
-      if (hasChanges) {
+      if (hasChanges()) {
         // Prevent the default unload, which tells the browser to show its native dialog.
```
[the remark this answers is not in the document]
molund: Made it accept either a boolean or a function.

ayumi-oxd (approved):
LGTM!

frontend/src/router/pages/advisories/advisory/Advisory.jsx
```diff
@@ -134,7 +134,7 @@ export default function Advisory({ mode }) {
   const isAuthenticated = auth.isAuthenticated;
 
   // Use a route blocker to show a confirmation dialog when the user attempts to navigate away with unsaved changes
-  const blocker = useBlocker(dataChanged);
+  const blocker = useBlocker(() => dataChanged.current);
```
duncan-oxd, line 137: Would this work to just pass in `dataChanged.current` as a ref value? Or does it have to be wrapped in a function?

frontend/src/router/pages/advisories/advisory/Advisory.jsx
```diff
@@ -168,10 +168,10 @@ export default function Advisory({ mode }) {
 
   // Set dataChanged to true when any of the form fields change, to enable the navigation guard
   const markChanged = useCallback(() => {
-    setDataChanged(true);
+    dataChanged.current = true;
   }, []);
 
-  useNavigationGuard(dataChanged);
+  useNavigationGuard(useCallback(() => dataChanged.current, []));
```
duncan-oxd, line 174: Same question about passing the ref value directly vs as a function here. I've got some reading to do!

duncan-oxd (approved):
Looks good to me, too! I had a couple questions about that hasChanges/dataChanged ref and callback pattern, but I'll look into it anyway 👍

frontend/src/router/pages/advisories/advisory/Advisory.jsx
molund: Copilot actually decided to use a function not me. But this is how CoPilot defended it...

```
useNavigationGuard(dataChanged.current);
```

would only pass whatever value existed on the last render, which could be stale.

Passing a function lets the hook check the ref’s current value when navigation actually happens.

PR #639, opened 2026-06-24
CMS-1821: Fix populating annual date ranges bug

ayumi-oxd:
### Jira Ticket

CMS-1821

### Description

- Fixed the populating annual date ranges bug: When the new season requested, the dates are populated but with incorrect end year information


molund:
Looks like it should fix the issue 👍

molund (approved):

PR #644, opened 2026-06-25
CMS-1834: Create edit published tab

ayumi-oxd:
### Jira Ticket

CMS-1834

### Description

- Renamed the "Edit and review" tab to "Submit" and updated related files accordingly.
- Added a new "Edit published" tab and corresponding API endpoint
  - The `edit-published` endpoint currently uses the `park` endpoint with the status set to "published", additional details will be implemented later
- Added `/edit-published` path


molund:
Minor naming nit: I was okay with "Submit" as a tab label, but now that the page itself has been been named SubmitPage, the standalone "Submit" file names feel a little odd to me.
Maybe rename:

- SubmitTable → SubmitPageTable
- submitFilters → submitPageFilters

This might just be showing my age as an old web developer, but whenever I see "Submit" by itself my brain immediately assumes it's an `<input type="submit" />` rather than a page name. 😄 Keeping SubmitPage in the component names makes the intent a bit clearer and keeps the naming consistent with the page.

molund (approved):

PR #645, opened 2026-06-26
CMS-1861: Display rec site link if it has isDisplayed

ayumi-oxd:
### Jira Ticket

CMS-1861

### Description

- Display the recreation site as a link when `isDisplayed` is true, otherwise, display it as plain text


frontend/src/components/advisories/composite/advisorySummaryView/AdvisorySummaryView.jsx
molund: The fallback should include the `recResourceId` if there is one.

ayumi-oxd: Good catch! Also, it doesn't need to check for `recResourceId`, since it's a required field and should always be present.

molund (approved):

PR #648, opened 2026-06-29
CMS-1839: DOOT bookmarkable form URLs

molund:
### Jira Ticket

CMS-1839

### Description
This change introduces unique, shareable URLs for each season form, enabling users to copy and share direct links to a specific Park, Area, or Feature.
The React application routing and form templates have been updated so that every season form is associated with its own URL using the seasonId of the record. URLs follow the format:
https://alpha-dev-staff.bcparks.ca/dates/edit/park-area/12345
where 12345 represents the seasonId of the ParkArea.
When users navigate between forms within the application, the browser URL updates automatically to reflect the currently open form. Users can also navigate directly to a specific form by entering or pasting one of these URLs into the browser address bar. When accessed directly, the application loads and automatically opens the corresponding edit form in the form panel.

The “Edit public dates on API?” and “Edit approved dates?” dialogs have been moved from EditAndReview.jsx to FormPanel.jsx. However, they now have a slightly different behaviour compared to before. Instead of blocking navigation to the form panel, the dialogs now load on top of it. If you click “Cancel” or "Close", the form panel will close along with the dialog. 

This pattern will be more flexible when we start using the form panel on other screens because it’s not tied to the data from the main table on the EditAndReview screen. 

Note: Support for other years will be added later as part of the ongoing "Edit Published Tab" development.

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -12,6 +12,7 @@ import {
   getDateTypesForFeature,
   getDateTypesForPark,
 } from "../../utils/dateTypesHelpers.js";
+import { checkSeasonUserAccess } from "../../utils/seasonHelpers.js";
 
 import {
   Park,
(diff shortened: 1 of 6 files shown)


ayumi-oxd (approved):
Looks great!

PR #651, opened 2026-07-01
CMS-1836: Add edit published table

ayumi-oxd:
### Jira Ticket

CMS-1836

### Description

- Added the Edit published table
  - Displayed past published seasons based on the selected park search
  - Added an Edit button to each season row
    - Allowed users to open and edit the most current past year season form
- Created `IconButton` component so it can be reused in the different tables


frontend/src/router/pages/EditPublishedPage.scss
```diff
@@ -0,0 +1,20 @@
...
+      &--park-area,
+      &--park-area-feature,
+      &--feature {
+        th,
+        td {
+          background-color: var(--bs-table-bg);
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
ayumi-oxd: This is fine.

frontend/src/router/pages/EditPublishedPage.scss
```diff
@@ -0,0 +1,20 @@
+.page.edit-published {
+  table {
+    tr.table-row {
+      &--park {
+        th,
+        td {
+          background-color: var(--surface-color-primary-disabled);
```
duncan-oxd: This isn't a disabled element technically but I don't know if there's a better name for this color in the glossary. Figma has it as the surface.color.secondary.button.hover, which is also not what this element is but it might be slightly safer to use in case the design tokens change some day.

frontend/src/router/pages/EditPublishedPage.jsx
```diff
@@ -23,6 +29,90 @@ export default function EditPublishedPage() {
...
+          (season) =>
+            season.seasonType === seasonType &&
+            season.operatingYear === targetOperatingYear,
+        ) || null
+      );
+    }
+
+    // Park-level seasons (regular and winter)
+    const regularSeason = getSeasonByYear(selectedPark.seasons, "regular");
+    const winterSeason = getSeasonByYear(selectedPark.seasons, "winter");
```
duncan-oxd: Let's use the constants/seasonType.js for these "winter" and "regular" strings

duncan-oxd (changes requested):
Cool! Less new code than I was expecting to implement this. I had a couple of very small change requests before we merge 👍

duncan-oxd (approved):

PR #654, opened 2026-07-03
CMS-1838: Disable form buttons until edits are made

ayumi-oxd:
### Jira Ticket

CMS-1838

### Description

- Disabled the edit published form buttons until edits are made

### DIFF
--- a/frontend/src/components/FormPanel.jsx
+++ b/frontend/src/components/FormPanel.jsx
@@ -60,13 +60,17 @@ function Buttons({
   approver,
   submitter,
   loading = false,
+  disableButtons = false,
 }) {
+  const isDisabled = loading || disableButtons;
+
   return (
     <div>
       <button
         type="button"
         onClick={onSave}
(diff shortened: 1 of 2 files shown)


frontend/src/components/FormPanel.jsx
```diff
@@ -456,6 +463,9 @@ function SeasonForm({
     return changesPayload.notes.length > 0;
   }, [season, changesPayload, apiData]);
 
+  // In the Edit published page, buttons should only be enabled after edits are made
+  const disableButtons = showOperatingYearSelect && !dataChanged;
```
duncan-oxd, line 467: We should double-check with UX if we want this to behavior on the regular edit form as well. It seems odd to have the buttons disabled on some forms but not others.

duncan-oxd (approved):

frontend/src/components/FormPanel.jsx
ayumi-oxd: Good point! I checked it with Amanda and she confirmed it's only for the Edit publish form.

PR #674, opened 2026-07-27
CMS-1877: Display winter fee dates on submit table

ayumi-oxd:
### Jira Ticket

CMS-1877

### Description

- Displayed the winter fee dates on the submit table for approvers
  - Displayed "(calculated)" next to "Winter fee"
  - Displayed the winter fee dates at bottom
- Split from CMS-1297


frontend/src/components/SubmitPageTable.jsx
```diff
@@ -122,27 +128,56 @@ function DateTableRow({ groupedDateRanges, currentYear }) {
...
-      </td>
-      <td>
-        <DateRangesList dateRanges={yearsObj[currentYear]} />
-      </td>
-    </tr>
-  ));
+  return sortedDateTypes.map(([dateTypeName, yearsObj]) => {
+    const dateTypeNumber = getDateTypeNumber(yearsObj);
+    const isCalculatedWinterFee =
+      showCalculatedWinterFeeLabel && dateTypeNumber === DATE_TYPE.WINTER_FEE;
```
duncan-oxd: I feel like we should make `showCalculatedWinterFeeLabel` more generic, maybe rename it to `showCalculatedDateTypeLabels` or something? 

We can leave the code the same: Just check WINTER_FEE because it's the only type that applies and I don't think we want to store that info in the DB, but the prop doesn't have to be specific to winter fees.

How does that sound?

frontend/src/components/SubmitPageTable.jsx
```diff
@@ -122,27 +128,56 @@ function DateTableRow({ groupedDateRanges, currentYear }) {
...
+    return groupedDateRanges;
+  }
+
+  // Hide winter fee dates for non-approvers
+  return Object.fromEntries(
+    Object.entries(groupedDateRanges).filter(([, yearsObj]) => {
+      const firstRange = Object.values(yearsObj || {}).find(
+        (rangesForYear) => Array.isArray(rangesForYear) && rangesForYear.length,
+      )?.[0];
+      const dateTypeNumber = firstRange?.dateType?.dateTypeNumber;
```
duncan-oxd: Is this doing the same thing as the `getDateTypeNumber` helper above? Could we reuse that here?

duncan-oxd (changes requested):
A few small requests!

duncan-oxd (approved):
🙌

PR #675, opened 2026-07-27
CMS-1870: Import dates modified in Strapi back into DOOT

molund:
### Jira Tickets

CMS-1870
CMS-1858

### Description
Reverse syncs operating year 2026+ dates that were modified in Strapi back into DOOT. Intended as a one-time operation in late August or early September 2026, before 2027 date collection begins.

Two scripts handle different entity types: `import-2026-strapi-feature-dates.js` syncs feature-level date ranges, and `import-2026-strapi-protected-area-dates.js` syncs park-level date ranges.

The scripts create `Publishable` and `Dateable` records if missing, create missing `Season` records for the operating year, sync `DateRange` record from JSON data by creating new records and deleting orphaned ones, and create or update `DateRangeAnnual` records based on the `is_date_annual` flag. All changes are wrapped in a transaction and rolled back on error.

Data files `feature-dates.json` and `protected-area-dates.json` are included in the task folder as JSON files rather than fetched from the Strapi REST API because the API does not expose `created_by_id` and `updated_by_id`, which are central to the sync logic.

Dry-run mode is available by setting `const dryRun = true;` at the top of either script to preview changes without writing to the database.

### DIFF
--- a/backend/tasks/import-2026-strapi-dates/PSQL.md
+++ b/backend/tasks/import-2026-strapi-dates/PSQL.md
@@ -1,0 +1,81 @@
+## Strapi data extraction
+
+Data is extracted from Strapi via SQL queries because the REST API does not expose `created_by_id` and `updated_by_id`. When data is added to Strapi through API calls, these fields are always `null`. Therefore, if records were added or edited by a human, these fields will be non-null.
+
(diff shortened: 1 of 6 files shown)


ayumi-oxd (approved):
Looks good!

PR #680, opened 2026-07-28
CMS-1935: Implement status resolution and required team review checks

duncan-oxd:
### Jira Ticket

CMS-1935

### Description

The first phase of the mutli-approval feature. Only backend and DB changes here:

- DB migration to add two new columns to the Season model: `reservationSvcApproved` & `informationSvcApproved`
- In the migration, update all the seasons that are already approved to have both fields "true" (everything else is both false)
- Add new groups in Keycloak as sub-groups within HQ Staff: DOOT Reservation Services approvers & DOOT Information Services approvers
- Added a role for the two new Keycloak groups: `doot-reservation-svc-approver` & `doot-information-svc-approver` respectively
- Painstakingly added all the HQ Staff people to the relevant sub-groups on all three Keycloak environments. (Everyone went to IS except 5 specific RS people, so our accounts are probably in IS)
- Renamed the existing `resolveNewSeasonStatus` helper function to `resolveSeasonCreationStatus` so it won't get confused with a similarly named function in this branch.

Added new functions in the season/save endpoint:
- Determine which team the user belongs to, if any
- If the user is approving a season: Update the individual IS or RS team approval flag
- Check if all of the required IS/RS team approval flags are true and set the overall status to APPROVED if applicable

Most things only need one team to approve, but a season can have a combination that would require both. (Reservation system AND a gate, or an area with several features)
image

There's a handy Confluence page with some examples, linked in the Figma for this ticket.

**_Be sure to run the DB migration!_**


backend/utils/seasonApprovalHelpers.js
```diff
@@ -0,0 +1,177 @@
...
+ * @returns {boolean} True when the gate state requires Information Services review
+ */
+export function requiresGateApproval(oldGateDetail, newGateDetail) {
+  const oldHasGate = oldGateDetail?.hasGate === true;
+  const newHasGate = newGateDetail?.hasGate === true;
+
+  // Information Services team must review seasons with gate info,
+  // or cases where a gate was removed.
+  return newHasGate || (oldHasGate && newGateDetail?.hasGate === false);
+}
```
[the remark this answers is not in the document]
duncan-oxd: The frontend _should_ always send the full form data, including hasGate, but we could guard against it and treat an undefined value as hasGate=false.

I'll change it to be more explicit:

```js
  const oldHasGate = oldGateDetail?.hasGate === true;
  const newHasGate = newGateDetail?.hasGate === true;
  const removedGate = oldHasGate && !newHasGate;

  return newHasGate || removedGate;
```

backend/routes/api/seasons.js
molund, line 739: This files is 1388 lines now 🤯

I think maybe we should split it up?

ayumi-oxd (approved):
Looks good! btw what is SVC?

backend/routes/api/seasons.js
duncan-oxd: We could move each endpoint into its own file at the very least!

duncan-oxd:
> Looks good! btw what is SVC?

It's just short for "service" 😅 I felt like I should abbreviate the long names somehow, but I didn't want to shorten "Information" or "Reservation" because that's the more important part!

backend/routes/api/seasons.js
duncan-oxd: I just pushed a commit to split all of the helper functions into files in /utils
The "utils" function can probably be renamed or re-organized some day but this cuts the file length in half, and the new file names make the stuff easy enough to find.

duncan-oxd:
I split the non-endpoint functions into separate files to cut the size of seasons.js in half. I also just pushed a small tweak to make sure the team-specific approval gets saved to the DB, even if it's not required. (If IS approval is required but RS is not, we'll still update `reservationSvcApproved` in the DB for record-keeping purposes, even though it won't affect the overall season status.

PR #683, opened 2026-07-29
CMS-1836: QA fixes to the "Edit published" results table

molund:
### Jira Ticket

CMS-1836

### Description

This PR addresses issues reported by QA:

1. Updated table rows to look and behave like BCGov secondary buttons.
2. Removed the default grey background from the "Tiers and Gate" row. The Figma design was showing a hover state rather than the intended default state.
3. Refined backend filtering for the "Edit Published" screen. The screen is intended to display all parks, park-areas and features with entries from the previous date collection period regardless of publishing status. The year calculation was also problematic because the "current date collection year" is arbitrarily determined by when the `tasks/create-seasons/create-seasons.js` script is run; it cannot be determined from the current calendar year alone.
4. Refactored IconButton so it would accept a `tabIndex={-1}` prop via a spread operator. Also removed an unused `textColor` prop from IconButton and a corresponding `color` prop from ApproveButton and StatusTableRow.  This prop wasn't being used and it was actually expecting a classname rather than a color string. It was replaced with extraClassName in IconButton which is less confusing.

### DIFF
--- a/backend/routes/api/edit-published.js
+++ b/backend/routes/api/edit-published.js
@@ -1,4 +1,5 @@
 import { Router } from "express";
+import asyncHandler from "express-async-handler";
 import { Season } from "../../models/index.js";
 import * as USER_ROLES from "../../constants/userRoles.js";
 import { checkPermissions } from "../../middleware/permissions.js";
@@ -11,7 +12,7 @@ const router = Router();
 router.get(
   "/",
   checkPermissions([USER_ROLES.APPROVER]),
-  async (req, res, next) => {
+  asyncHandler(async (req, res, next) => {
     // get the max season (latest operating year) from the db
     const maxSeason = await Season.findOne({
       order: [["operatingYear", "DESC"]],
@@ -30,8 +31,9 @@ router.get(
       operatingYear: previousDateCollectionYear,
     };
 
-    parkRoutes.handle(req, res, next);
-  },
+    return next();
+  }),
+  parkRoutes,
 );
 
 export default router;
--- a/backend/routes/api/edit-published.js
+++ b/backend/routes/api/edit-published.js
@@ -1,23 +1,37 @@
 import { Router } from "express";
-import * as SEASON_STATUS from "../../constants/seasonStatus.js";
+import { Season } from "../../models/index.js";
 import * as USER_ROLES from "../../constants/userRoles.js";
 import { checkPermissions } from "../../middleware/permissions.js";
 import parkRoutes from "./parks.js";
 
 const router = Router();
 
-// Reuse parkRoutes for the Edit published page
-// Get all parks with published seasons for the previous year
-router.get("/", checkPermissions([USER_ROLES.APPROVER]), (req, res, next) => {
-  const minOperatingYear = new Date().getFullYear() - 1;
+// Reuse parkRoutes for the previous-year edit page.
+// Get all parks with seasons for the previous year, regardless of status or seasonType.
+router.get(
+  "/",
+  checkPermissions([USER_ROLES.APPROVER]),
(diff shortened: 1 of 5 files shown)


backend/routes/api/edit-published.js
```diff
@@ -1,23 +1,35 @@
...
+    // operatingYear in the database is one year ahead of the active camping season.
+    const campingDateCollectionYear = maxSeason.operatingYear - 1;
+    const previousDateCollectionYear = campingDateCollectionYear - 1;
 
-  parkRoutes.handle(req, res, next);
-});
+    req.query = {
+      ...req.query,
+      operatingYear: previousDateCollectionYear,
+    };
```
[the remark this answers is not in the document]
molund: Deriving the year from the DB will handle the autumn collection scenario where the previous collection year is actually the current calendar year. There are no real-world scenarios that will be broken by the clamping so I'm going to leave it as-is.

backend/routes/api/edit-published.js
```diff
@@ -1,23 +1,37 @@
...
-// Reuse parkRoutes for the Edit published page
-// Get all parks with published seasons for the previous year
-router.get("/", checkPermissions([USER_ROLES.APPROVER]), (req, res, next) => {
-  const minOperatingYear = new Date().getFullYear() - 1;
+// Reuse parkRoutes for the previous-year edit page.
+// Get all parks with seasons for the previous year, regardless of status or seasonType.
+router.get(
+  "/",
+  checkPermissions([USER_ROLES.APPROVER]),
+  async (req, res, next) => {
```
duncan-oxd: To make API route handlers async, we need to use the "express-async-handler" wrapper. You can just copy the pattern from one of the other API route files, like parks.js 👍 

The wrapper makes sure async errors get forwarded to to Express' error handlers. I think this version of Express existed before Javascript had async/await 😆

backend/routes/api/edit-published.js
```diff
@@ -1,23 +1,37 @@
...
-  parkRoutes.handle(req, res, next);
-});
+    req.query = {
+      ...req.query,
+      operatingYear: previousDateCollectionYear,
+    };
+
+    parkRoutes.handle(req, res, next);
+  },
+);
```
duncan-oxd, line 37: I think this is fine as written, but it might be more _Express-ish idiomatic_ if we pass the parkRoutes as the next handler after this one (ie, the next argument for `.get`):

```suggestion

    // Delegate the response to the /parks/ GET handler
    return next();
  },
  
  parkRoutes,
);
```

Express.js middleware composition is pretty much just a stack of handlers connected by `next` functions. `parkRoutes` is middleware too, so Express will continue routing through that router's matching handlers ("/")

```js
router.get("/", thisHandler, otherHandler, somethingElse)
```

duncan-oxd (dismissed):
It's the end of the day and I wasn't able to get anything to appear in my "Edit published" tab, but I have a few thoughts about the API route handler in the meantime. The frontend code changes look great 👍

backend/routes/api/edit-published.js
molund: I updated this any it works now, but I'm still a bit vague on how the express chaining works

backend/routes/api/edit-published.js
molund: added the wrapper

molund:
> It's the end of the day and I wasn't able to get anything to appear in my "Edit published" tab, but I have a few thoughts about the API route handler in the meantime. The frontend code changes look great 👍

This is now fixed with the latest commit

ayumi-oxd (approved):
LGTM!

PR #687, opened 2026-07-31
CMS-1878: Hide winter fee dates when feature.hasWinterFeeDates is false

molund:
### Jira Ticket

CMS-1878

### Description
This change hides the calculated winter fee date row from the table when `feature.hasWinterFeeDates`. The existence of this row is confusing because it is essentially hardcoded in these scenarios with no way to force recalculation.
This scenario typically arises from data that was originally imported from Strapi.

### DIFF
--- a/backend/routes/api/parks.js
+++ b/backend/routes/api/parks.js
@@ -1,4 +1,4 @@
-import { Op } from "sequelize";
+import { Op, col, where as sqlWhere } from "sequelize";
 import { Router } from "express";
 import _ from "lodash";
 import sequelize from "../../db/connection.js";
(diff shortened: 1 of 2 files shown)


backend/routes/api/parks.js
```diff
@@ -75,6 +75,30 @@ function seasonModel(minYear, required = true, seasonStatus = null) {
...
+    ...seasonModel(minYear, true, seasonStatus),
+    where: {
+      operatingYear: {
+        [Op.gte]: minYear,
+      },
+      ...(seasonStatus ? { status: seasonStatus } : {}),
+      // SQL-level guard: if a feature has no winter fee dates, exclude winter seasons.
+      [Op.or]: [
+        { seasonType: { [Op.ne]: SEASON_TYPE.WINTER } },
+        sqlWhere(col("features.hasWinterFeeDates"), true),
```
ayumi-oxd, line 96: Good to know this exsists!

ayumi-oxd (approved):

backend/routes/api/parks.js
duncan-oxd: Wow! I didn't know about this. I tried to do similar things in the past with a bunch of funky syntax but it never worked right and I had to fall back to separate queries. 📝 👀

PR #689, opened 2026-07-31
CMS-1838: Update past seasons option for group camping feature

ayumi-oxd:
### Jira Ticket

CMS-1838

### Description

- In the Edit published, added the current operating year option for those feature types (Group campground and Picnic shelter) as they always carry one extra season ahead

### DIFF
--- a/backend/routes/api/seasons.js
+++ b/backend/routes/api/seasons.js
@@ -154,6 +154,7 @@ router.get(
       dateTypes: orderedDateTypes,
       icon: seasonModel.feature.featureType.icon,
       featureTypeName: seasonModel.feature.featureType.name,
+      featureTypeNumber: seasonModel.feature.featureType.featureTypeNumber,
       name: seasonModel.feature.name,
       parkName: seasonModel.feature.park.name,
       parkWinterDates,
@@ -252,6 +253,7 @@ router.get(
 
     let icon = null;
     let featureTypeName = null;
+    let featureTypeNumber = null;
 
     // If there are features in the Park Area, use the first feature's type
     if (seasonModel.parkArea.features.length > 0) {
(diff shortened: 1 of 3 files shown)


backend/routes/api/seasons.js
```diff
@@ -481,7 +482,7 @@ router.get(
       where: {
         publishableId: currentSeason.publishableId,
         seasonType: currentSeason.seasonType,
-        status: STATUS.PUBLISHED,
+        [Op.or]: [{ status: STATUS.PUBLISHED }, { operatingYear: currentYear }],
       },
       order: [["operatingYear", "DESC"]],
```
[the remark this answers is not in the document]
ayumi-oxd: This is fine since the past seasons are published in general.

backend/routes/api/seasons.js
```diff
@@ -468,6 +473,7 @@ router.get(
   "/options/:seasonId",
   asyncHandler(async (req, res) => {
     const seasonId = Number(req.params.seasonId);
+    const currentYear = new Date().getFullYear();
```
molund, line 476: Using the calender year in DOOT operatingYear logic is a bit problematic. New seasons get created in the fall but the calendar year logic only recognizes Jan 1 as the next year.

Look a the way it's done here: https://github.com/bcgov/bcparks-staff-portal/blob/bb4d097638e0afe46a7a29e75d4618b9878bb588/backend/routes/api/edit-published.js#L17

backend/routes/api/seasons.js
```diff
@@ -481,7 +487,7 @@ router.get(
       where: {
         publishableId: currentSeason.publishableId,
         seasonType: currentSeason.seasonType,
-        status: STATUS.PUBLISHED,
+        [Op.or]: [{ status: STATUS.PUBLISHED }, { operatingYear: currentYear }],
```
molund: I don't think the `status: STATUS.PUBLISHED` filter is necessay.

You could just filter based on operatingYear

```
  where: {
    publishableId: currentSeason.publishableId,
    seasonType: currentSeason.seasonType,
    status: STATUS.PUBLISHED,
    operatingYear: {
      [Op.lte]: currentYear,
    },
```

PR #692, opened 2026-08-03
CMS-1940: Add pending review subsets

ayumi-oxd:
### Jira Ticket

CMS-1940

### Description

- Added "IS review" and "RS review" below "Pending HQ review" and have both indented as a subset of the HQ review  
  - Only visible to the approver roles
  - When "Pending HQ review" is checked, both "IS review" and "RS review" is checked
  - When only one "IS review/RS review" is checked, "Pending HQ review" should still be checked


duncan-oxd:
Rebasing onto `main` to resolve a merge conflict I created when I merged #697 👍

frontend/src/components/FilterStatus.jsx
```diff
@@ -75,6 +76,36 @@ export default function FilterStatus({
         label: `Status: ${labelByValue[status]}`,
 
         remove(filters) {
+          const HQ = STATUS.PENDING_REVIEW.value;
+          const IS = STATUS.IS_REVIEW_FILTER.value;
+          const RS = STATUS.RS_REVIEW_FILTER.value;
```
duncan-oxd: I'd prefer to keep the `PENDING_REVIEW` / `IS_REVIEW` / `RS_REVIEW` variable names for clarity instead of the 2-letter variable names unless there's an important reason to shorten them

frontend/src/lib/submitPageFilters.js
```diff
@@ -1,4 +1,45 @@
 import * as DATE_TYPES from "@/constants/dateType.js";
+import * as STATUS from "@/constants/seasonStatus.js";
+
+function matchesStatusFilters(season, statusFilters) {
+  if (!season || !statusFilters.length) return false;
+
+  const hasParentPending = statusFilters.includes(STATUS.PENDING_REVIEW.value);
+  const hasIsReview = statusFilters.includes(STATUS.IS_REVIEW_FILTER.value);
+  const hasRsReview = statusFilters.includes(STATUS.RS_REVIEW_FILTER.value);
```
duncan-oxd: These variable names confused me at first; "has IS review" means "the status filters list has the IS review filter" but it looks like "we want to filter for seasons that have been reviewed by the IS team"

I don't know if there's a way we can re-word this to make it super obvious, but I get it now 😅

frontend/src/components/FilterStatus.jsx
```diff
@@ -75,6 +76,36 @@ export default function FilterStatus({
...
+
+          // Keep badge-removal behavior consistent with the dropdown's
+          // parent/child status-selection rules.
+          if (status === HQ) {
+            updateFilter(
+              "status",
+              filters.status.filter(
+                (value) => value !== HQ && value !== IS && value !== RS,
+              ),
```
duncan-oxd: We're using `without` elsewhere in this file, and this might be the best candidate for it because it's filtering 3 different things. I think we could change this to:
```suggestion
              without(filters.status, HQ, IS, RS)
```

It's easy to read either way, but it's an option if we're already using it in this file/function.

frontend/src/lib/submitPageFilters.js
```diff
@@ -1,4 +1,45 @@
...
+  }
+
+  // Treat "review" as "still awaiting that team's approval".
+  if (hasIsReview && season.informationSvcApproved !== true) {
+    return true;
+  }
+
+  if (hasRsReview && season.reservationSvcApproved !== true) {
+    return true;
+  }
```
duncan-oxd, line 57: _Updated this comment to remove confusing stuff and reflect the new strategy!_

Just because the `_Approved` column is false doesn't mean it _needs_ that team's approval to be approved and published. There's now a [set of utility functions](https://github.com/bcgov/bcparks-staff-portal/blob/main/backend/utils/seasonApprovalHelpers.js) you can use to determine which approvals are needed, based on the season data. 

I recently calculated that same thing for the CSV export in #702 so you can use the same patterns here; refer to the `seasonRequiresInformationSvcApproval` and `seasonRequiresReservationSvcApproval` functions and pretty much copy those patterns. The CSV endpoint gets _all_ the season changelogs but you really only need _one_ where the gate status changes from true to false.

I think we'll need to calculate "requires IS approval?" and "requires RS approval" for every current season in the `/parks` or `/parks/metadata` endpoint so the filters can use them. If it slows down the query a lot, we'll want to load that in the metadata endpoint. The metadata endpoint currently only loads park-level stuff and not season-level stuff so the function to merge the data together might need some tweaking. Let me know if you have questions about that.

Note that you won't be using `getRequiredApprovalsForSeason` because that function is used in the `/save` endpoint and it requires old data and new data. That function has a misleading name, sorry! In your case you only need to look at the current data in the DB, same as in the CSV export.

frontend/src/router/pages/SubmitPage.jsx
```diff
@@ -101,12 +103,42 @@ function SubmitPage() {
...
+      { value: STATUS.REQUESTED.value, label: STATUS.REQUESTED.label },
+      {
+        value: STATUS.PENDING_REVIEW.value,
+        label: STATUS.PENDING_REVIEW.label,
+      },
+      { value: STATUS.APPROVED.value, label: STATUS.APPROVED.label },
+      { value: STATUS.PUBLISHED.value, label: STATUS.PUBLISHED.label },
+    ];
+
+    if (isApprover) {
```
duncan-oxd: We could use a few `.push` lines in a row, or a conditional spread like:
```js
options = [ stuff, ...(isAppprover ? approverStuff : []), otherStuff ];
```
Then we wouldn't need to do a `.find` in the array we're building.

backend/utils/seasonApprovalHelpers.js
```diff
@@ -66,6 +66,125 @@ export async function hasGateRemoved(seasonId) {
...
+
+  return false;
+}
+
+/**
+ * Returns whether Reservation Services team approval is required for a season.
+ * @param {Season} season Season with park/parkArea/feature associations
+ * @returns {boolean} True when RS team approval is required
+ */
+export function seasonRequiresReservationSvcApproval(season) {
```
duncan-oxd, line 118: If `seasonRequiresInformationSvcApproval` and `seasonRequiresReservationSvcApproval` are exact copies of what's added to export.js in #702 then we'll end up with duplicated code when we merge. Let's remember to rebase one branch onto the other and `import` these functions into export.js from the helpers file before we merge it 👍

backend/utils/seasonApprovalHelpers.js
```diff
@@ -66,6 +66,125 @@ export async function hasGateRemoved(seasonId) {
...
+      annotateCurrentSeason(parkArea.currentSeason?.winter, parkAreaContext);
+    });
+
+    park.features.forEach((feature) => {
+      const featureContext = {
+        feature: { inReservationSystem: feature.inReservationSystem },
+      };
+
+      annotateCurrentSeason(feature.currentSeason?.regular, featureContext);
+      annotateCurrentSeason(feature.currentSeason?.winter, featureContext);
```
duncan-oxd: Are the winter seasons for Features and ParkAreas relevant? Or do we only need to do this for the regular seasons at those levels? (I forget where/how that data is used in the frontend 😆)

backend/utils/seasonApprovalHelpers.js
```diff
@@ -66,6 +66,125 @@ export async function hasGateRemoved(seasonId) {
...
+        : [],
+    };
+
+    currentSeason.requiresInformationSvcApproval =
+      seasonRequiresInformationSvcApproval(seasonContext);
+    currentSeason.requiresReservationSvcApproval =
+      seasonRequiresReservationSvcApproval(seasonContext);
+  }
+
+  parks.forEach((park) => {
```
duncan-oxd: Since we're changing the parks object and the stuff inside it by reference, and then returning it anyway, let's change this to a `.map` instead of a `.forEach` just so it's more obvious that we're going to change it and return it 👍

duncan-oxd (approved):
Looks great! I had one request about changing a .forEach to a .map for clarity, and a note for both of us about rebasing when we merge. Otherwise it looks good!

I think the Sept 24 stuff will be okay to merge later this week 👍

backend/utils/seasonApprovalHelpers.js
ayumi-oxd: Good catch! Winter fee dates at the feature level doesn't need to be "approved" since they are automatically calculated and approved.

